PR #151, opened 2016-12-11
Dht cas merge

ianopolous:
This PR merges the old DHT interface into the ContentAddressedStorage interface. In the process it matches our http protocol with IPFS's. It also moves the btree from a server side implementation to a client side one. And introduces caching for core node IPNS lookups. These are breaking changes for the data formats. They will also break again once I move to the ipfs dag api, which is cleaner than the object api which is now deprecated.


src/peergos/server/net/DHTHandler.java
```diff
@@ -1,120 +1,174 @@
...
+    }
+
+    private Map<String, List<String>> parseQuery(String query) {
+        if (query.startsWith("?"))
+            query = query.substring(1);
+        String[] parts = query.split("&");
+        Map<String, List<String>> res = new HashMap<>();
+        for (String part: parts) {
+            int sep = part.indexOf("=");
+            String key = part.substring(0, sep);
```
cboddy: key and  value should be url-decoded.

cboddy: Also this  could be  abstracted into a tested utility lib.

src/peergos/server/net/DHTHandler.java
```diff
@@ -1,120 +1,174 @@
...
-                    SortedMap<String, Multihash> namedLinks = new TreeMap<>();
-                    for (int i = 0; i < hashes.size(); i++)
-                        namedLinks.put(Integer.toString(i), hashes.get(i));
-                    MerkleNode obj = new MerkleNode(value, namedLinks);
-                    byte[] put = dht.put(obj).toBytes();
-                    new PutSuccess(httpExchange).accept(Optional.of(put));
-                } else {
+            switch (path) {
+                case "object/new":{
+                    UserPublicKey writer = UserPublicKey.fromString(last.apply("writer"));
```
cboddy: This switch makes the function a bit large. Perhaps each case could be  be  handled by a separate method?

src/peergos/server/net/DHTHandler.java
```diff
@@ -1,120 +1,174 @@
...
+                    dht._new(writer).thenAccept(newHash -> {
+                        Map res = new HashMap();
```
cboddy: Could this be typed?

src/peergos/server/net/DHTHandler.java
```diff
@@ -1,120 +1,174 @@
...
-                    byte[] value = Serialize.deserializeByteArray(din, ContentAddressedStorage.MAX_OBJECT_LENGTH);
-                    byte[] owner = Serialize.deserializeByteArray(din, UserPublicKey.MAX_SIZE);
-                    // TODO check we care about owner
-                    List<Multihash> hashes = new ArrayList<>();
-                    int nlinks = din.readInt();
-                    for (int i = 0; i < nlinks; i++)
-                        hashes.add(new Multihash(Serialize.deserializeByteArray(din, 1024)));
+            if (! path.startsWith(apiPrefix))
+                throw new IllegalStateException("Unsupported api version, required: " + apiPrefix);
+            path = path.substring(apiPrefix.length());
```
cboddy: It would be nice  to abstract this  logic into a re-usable super-class, something like a VersionedHTTPHandler.

src/peergos/server/net/DHTHandler.java
```diff
@@ -1,120 +1,174 @@
...
+                    break;
+                }
+                case "object/patch/add-link":{
+                    UserPublicKey writer = UserPublicKey.fromString(last.apply("writer"));
+                    Multihash hash = Multihash.fromBase58(args.get(0));
+                    String label = args.get(1);
+                    Multihash targetHash = Multihash.fromBase58(args.get(2));
+                    dht.addLink(writer, hash, label, targetHash)
+                            .thenAccept(resultHash -> {
+                                Map res = new HashMap();
```
cboddy: Could this be  typed?

src/peergos/server/net/DHTHandler.java
```diff
@@ -1,120 +1,174 @@
...
+                    break;
+                }
+                default: {
                     httpExchange.sendResponseHeaders(404, 0);
                 }
             }
         } catch (Exception e) {
             e.printStackTrace();
-            throw e;
+            throw new RuntimeException(e);
```
cboddy: It would be nice to  send  back an internal-server-error to the client here.

src/peergos/server/net/MultipartReceiver.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class MultipartReceiver {
+
+    public static byte[] extractFile(InputStream in, String boundary) {
+        try {
+            String first = readLine(in);
+            if (!first.substring(2).equals(boundary))
+                throw new IllegalStateException("Incorrect boundary! " + boundary + " != " + first.substring(2));
+            String filenameLine = readLine(in);
+            String contentType = readLine(in);
```
cboddy: This looks  super useful  and re-usable. Could we add a test for the basic functionality and error conditions?

src/peergos/server/net/MultipartReceiver.java
```diff
@@ -0,0 +1,46 @@
...
+                file.write(buffer, 0, r);
+            byte[] bytes = file.toByteArray();
+            // 8 is two \r\n and two -- at end of stream
+            return Arrays.copyOfRange(bytes, 0, bytes.length - (8 + boundary.length()));
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static String readLine(InputStream in) throws IOException {
```
cboddy: Perhaps this  function should  take a  max-length parameter like the  Serialize.deserializeString method to stop reading an  infinitely long stream?

src/peergos/server/storage/CachingStorage.java
```diff
@@ -1,43 +1,68 @@
...
         this.target = target;
-        this.cache = new LRUCache(cacheSize);
+        this.cache = new LRUCache<>(cacheSize);
         this.maxValueSize = maxValueSize;
     }
 
     @Override
-    public Multihash put(MerkleNode object) {
-        return target.put(object);
+    public CompletableFuture<Multihash> _new(UserPublicKey writer) {
```
cboddy: Could we  call this something more descriptive like `createEmpty`?

src/peergos/server/storage/IpfsDHT.java
```diff
@@ -27,54 +30,94 @@ public IpfsDHT() {
     }
 
     @Override
-    public byte[] get(Multihash key) {
+    public CompletableFuture<Multihash> _new(UserPublicKey writer) {
```
cboddy: Again, could we call this something else?

src/peergos/server/storage/RAMStorage.java
```diff
@@ -1,27 +1,64 @@
...
 import java.util.*;
+import java.util.concurrent.*;
 
 public class RAMStorage implements ContentAddressedStorage {
     private Map<Multihash, byte[]> storage = new HashMap<>();
 
+    private final Set<Multihash> pinnedRoots = new HashSet<>();
+
+    @Override
+    public CompletableFuture<Multihash> _new(UserPublicKey writer) {
```
cboddy: Not a big fan of this name.

src/peergos/server/tests/MerkleNodeSerialization.java
```diff
@@ -0,0 +1,26 @@
...
+import peergos.shared.merklebtree.*;
+
+import java.io.*;
+import java.util.*;
+
+public class MerkleNodeSerialization {
+
+    @Test
+    public void inverses() throws IOException {
+        MerkleNode original = new MerkleNode(new byte[0], Collections.emptyList());
```
cboddy: Could we add a test for a non empty MerkleNode?

src/peergos/server/storage/CachingStorage.java
ianopolous: this is just matching the ipfs name which is new

src/peergos/server/storage/IpfsDHT.java
ianopolous: same as ipfs

src/peergos/server/storage/RAMStorage.java
ianopolous: just matching ipfs

src/peergos/server/tests/MerkleNodeSerialization.java
ianopolous: done

src/peergos/server/net/DHTHandler.java
ianopolous: I looked into this, and because of the Java api requiring a HttpExchange this wouldn't actually encapsulate anything, The httpExchange contains the full path anyway. In essence to do this would require writing our own web framework, so basically jet all over again. So I think it's simpler how it is.

src/peergos/server/net/DHTHandler.java
ianopolous: They already are url decoded. URI.getQuery() does this

src/peergos/server/net/MultipartReceiver.java
ianopolous: this is already tested by all the user tests

src/peergos/server/net/DHTHandler.java
ianopolous: done

src/peergos/server/net/DHTHandler.java
ianopolous: it is json and it will change when we move to the dag api

src/peergos/server/tests/MerkleNodeSerialization.java
ianopolous: done

src/peergos/server/storage/RAMStorage.java
ianopolous: ok. done :-)

PR #153, opened 2016-12-26
Future proofing framework appears to be working!

ianopolous:
Moved to using WriterData which has extra properties which can be used for future proofing and also use cbor serialization.


ianopolous:
There appears to be a problem after signing up, creating a dir in your root fails to subsequently retrieve the subdir. If you then log out and in, then it is there. Also if you log out and in before trying to create it it works fine. So something is wrong with the UserContext returned from signUp.

src/peergos/shared/crypto/UserPublicKey.java
```diff
@@ -48,6 +49,17 @@ public static UserPublicKey fromPublicKeys(byte[] raw) {
...
     }
 
+    @Override
+    public CborObject toCbor() {
+        return new CborObject.CborByteArray(serialize());
+    }
+
+    public static UserPublicKey fromCbor(CborObject cbor) {
+        if (! (cbor instanceof CborObject.CborByteArray))
+            throw new IllegalStateException("UserPublicKEy cbor must be a byte[]! " + cbor);
```
cboddy: typo

src/peergos/shared/user/UserUtil.java
```diff
@@ -11,9 +11,15 @@
 
 public class UserUtil {
 
-    public static CompletableFuture<UserWithRoot> generateUser(String username, String password, LoginHasher hasher,
-                                            Salsa20Poly1305 provider, SafeRandom random, Ed25519 signer, Curve25519 boxer) {
-    	CompletableFuture<byte[]> fut = hasher.hashToKeyBytes(username, password);
+    public static CompletableFuture<UserWithRoot> generateUser(String username,
+															   String password,
```
cboddy: formatting?

PR #154, opened 2016-12-31
Made followers behave more intuitively. Someone doesn't show up as a …

ianopolous:
…follower now until they accept your follow request. (Only required a small change in protocol)

### DIFF
--- a/src/peergos/shared/user/SocialState.java
+++ b/src/peergos/shared/user/SocialState.java
@@ -9,17 +9,17 @@ import java.util.stream.*;
 @JsType
 public class SocialState {
 
-    public final List<FollowRequest> pending;
+    public final List<FollowRequest> pendingIncoming;
     public final Map<String, FileTreeNode> followerRoots;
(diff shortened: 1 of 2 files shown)


ianopolous:
Note that this is a PR into the future_proof branch (to make reviewing easier) rather than master.
So merge this PR before that one.

src/peergos/shared/user/SocialState.java
```diff
@@ -4,17 +4,30 @@
...
 import java.util.*;
+import java.util.stream.*;
 
 @JsType
 public class SocialState {
 
     public final List<FollowRequest> pending;
     public final Map<String, FileTreeNode> followerRoots;
     public final Set<FileTreeNode> followingRoots;
+    public final Map<String, FileTreeNode> pendingFollowRequestsToOthers;
```
cboddy: is the ToOthers specifying anything particular or could  is just be named ``pendingFollowRequests``?

ianopolous: done

PR #156, opened 2017-01-07
Split userpublickey

ianopolous:
Move boxing key from UserPublicKey WriterData. This means it is now automatically authenticated by the signing key.


ianopolous:
Locations now cannot include private keys. A separate SigningKeyPair must be explicitly present to commit a write.

PR #161, opened 2017-01-11
Remove erasure coding

cboddy:


src/peergos/shared/user/fs/FileUploader.java
```diff
@@ -38,8 +38,7 @@ public FileUploader(String name, AsyncReader fileData, int offsetHi, int offsetL
             this.props = fileProperties;
         if (baseKey == null) baseKey = SymmetricKey.random();
 
-        fragmenter = nAllowedFalures == 0 ?
-                new peergos.shared.user.fs.SplitFragmenter() : new peergos.shared.user.fs.ErasureFragmenter(nOriginalFragments, nAllowedFalures);
+        fragmenter = Fragmenter.getInstance();
```
ianopolous: I think we want to keep this controllable by the consumer

src/peergos/shared/user/fs/FileUploader.java
```diff
@@ -104,7 +103,7 @@ public FileUploader(String name, AsyncReader fileData, long offset, long length,
 
         peergos.shared.user.fs.Fragmenter fragmenter = nAllowedFalures == 0 ?
                 new peergos.shared.user.fs.SplitFragmenter() :
-                new peergos.shared.user.fs.ErasureFragmenter(nOriginalFragments, nAllowedFalures);
+                Fragmenter.getInstance();
```
ianopolous: same as above

PR #191, opened 2017-02-20
Added two method to ContentAddressedStorage to allow determination of…

ianopolous:
… space usage without downloading any blocks.

### DIFF
--- a/src/peergos/server/net/DHTHandler.java
+++ b/src/peergos/server/net/DHTHandler.java
@@ -105,6 +105,26 @@ public class DHTHandler implements HttpHandler
                     }).exceptionally(Futures::logError);
                     break;
                 }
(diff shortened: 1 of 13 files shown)


src/peergos/server/net/DHTHandler.java
```diff
@@ -117,8 +137,12 @@ public void handle(HttpExchange httpExchange) throws IOException {
     }
 
     private static Object wrapHash(Multihash h) {
+        return wrapHash("Hash", h);
+    }
+
+    private static Object wrapHash(String key, Multihash h) {
```
cboddy: Mind changing the type  of this to ``Map<String, Object>``

ianopolous: done

PR #207, opened 2017-04-15
Align the length of a split file to 256 bytes.

cboddy:


src/peergos/shared/user/fs/SplitFragmenter.java
```diff
@@ -12,11 +12,13 @@ public double storageIncreaseFactor() {
     }
 
     public byte[][] split(byte[] input) {
+        //align to 256 bytes
+        int len = input.length + (256 - input.length % 256);
```
ianopolous: what if input.length is a multiple of 256? wont this add an extra 256 bytes?

cboddy: It would get an extra 256 bytes of padding, fixed in the next commit.

src/peergos/shared/user/fs/SplitFragmenter.java
```diff
@@ -12,11 +12,18 @@ public double storageIncreaseFactor() {
     }
 
     public byte[][] split(byte[] input) {
+        //calculate padding length to align to 256 bytes
+        int padding = 0;
+        int mod = input.length % 256;
+        if (mod == 0 && input.length > 0)
+            padding = 256 - mod;
+        //align to 256 bytes
```
ianopolous: doesn't this mean padding is always 0, unless length is already aligned, when it is 256?

src/peergos/server/tests/FragmenterTest.java
```diff
@@ -57,7 +57,7 @@ private void test(byte[] input) throws IOException {
 
         for (byte[] bytes : split) {
             int length = bytes.length;
-            assertTrue(length > 0);
+            assertTrue(length >= 0);
```
ianopolous: In case is this zero?

PR #208, opened 2017-04-16
Separated out MutablePointers interface from Corenode interface.

ianopolous:
The former will be high volume and decentralized, the latter,
low volume and requiring consensus.


src/peergos/server/corenode/PinningMutablePointers.java
```diff
@@ -0,0 +1,63 @@
...
+
+    @Override
+    public CompletableFuture<Boolean> setPointer(PublicSigningKey owner, PublicSigningKey signer, byte[] sharingKeySignedBtreeRootHashes) {
+        // first pin new root
+        byte[] message = signer.unsignMessage(sharingKeySignedBtreeRootHashes);
+        HashCasPair cas = HashCasPair.fromCbor(CborObject.fromByteArray(message));
+        long t1 = System.currentTimeMillis();
+        return storage.recursivePin(cas.updated.get()).thenCompose(pins -> {
+            if (!pins.contains(cas.updated.get()))
+                return CompletableFuture.completedFuture(false);
```
cboddy: I'd suggest making this error state throw an error since it shouldn't happen.

src/peergos/server/mutable/HttpMutablePointerServer.java
```diff
@@ -0,0 +1,125 @@
...
+                }
+
+                dout.flush();
+                dout.close();
+                byte[] b = bout.toByteArray();
+                exchange.sendResponseHeaders(200, b.length);
+                exchange.getResponseBody().write(b);
+            } catch (Exception e) {
+                e.printStackTrace();
+                exchange.sendResponseHeaders(400, 0);
```
cboddy: Suggest sending back some message with the 400 response, ``e.getMessage()``?

PR #226, opened 2017-05-24
Make leaf nodes (file fragments) use raw format rather than cbor.

ianopolous:
This means that pins will be MUCH faster, and we also save a few bytes.

This is a breaking change. So I won't merge/deploy this until I can batch it with a few other breaking changes.


ianopolous:
This is no longer a breaking change, and results is 2X faster writes. :-)

PR #231, opened 2017-06-09
Keep track of expired username claims in tofu keystore.

ianopolous:
Auto renew username on login if it has expired (this need more testing)
Separate keychain validity from expiry.

This needs a solid review. Particularly for the security implications.


ianopolous:
I need to make tofu key store reverse lookup also use expired chains. And make get username handle public keys we don't have the username for.

PR #234, opened 2017-06-23
Move to using hashes of public keys instead of inband keys. (Post-quantum-crypto preparation)

ianopolous:
I'm struggling to fix whatever is stopping tests passing. This simplest that fails is RamUserTests.randomSignup()


ianopolous:
This is mostly good to go. All the tests except for the changePassword ones are passing

ianopolous:
All IPFS tests are passing locally with ipfs/master

src/peergos/server/corenode/JDBCCoreNode.java
```diff
@@ -326,10 +328,12 @@ public MetadataBlob selectOne()
     }
 
     private volatile boolean isClosed;
+    private final ContentAddressedStorage ipfs;
 
-    public JDBCCoreNode(Connection conn) throws SQLException
+    public JDBCCoreNode(Connection conn, ContentAddressedStorage ipfs) throws SQLException
```
ianopolous: Solid idea.

PR #246, opened 2017-08-23
username_validation: Added rule-set for valid usernames as enforced by JDBCCoreNode.

cboddy:


ianopolous:
looks like gwtc failed. there is nor Regex in gwt. So you'll need to do something simpler.

PR #249, opened 2017-08-31
implement thumbnails in java

kevodwyer:


kevodwyer:
Can you test with FUSE to confirm? not possible on my machine.

ianopolous:
You can test with the uploader class.

ianopolous:
It would be great if we could move the java impl to the server package and make it implement an interface, then we wouldn't need to dummy gwt implementations.

ianopolous:
Would also be good to have a test that generated a thumbnail in java, maybe for the logo or something that is already in git.

PR #253, opened 2017-09-23
update http headers for more safety

ianopolous:


ianopolous:
From: https://securityheaders.io/?q=demo.peergos.net&followRedirects=on

PR #256, opened 2017-09-30
Implement mime-type in FileProperties

ianopolous:
(and change the properties to cbor) It is calculated based on the file start. 

This will allow an easy solution to stop the video/audio player having to guess the mime type from the filename.


ianopolous:
note this is unfortunately a breaking change, but moving to cbor should make it much easier to modify the structure of file properties in the future

PR #266, opened 2017-11-08
unsharing

kevodwyer:


src/peergos/gwt/emu/java/nio/file/Path.java
```diff
@@ -4,31 +4,80 @@
...
+    {
+        test();
+    }
+    private static void assertTrue(boolean value, String msg) {
+        if(!value) {
+            throw new Error(msg);
+        }
+    }
+
+    private static void test(){
```
ianopolous: is this needed?

kevodwyer: I will remove the test method

PR #268, opened 2017-11-11
Blacklist

ianopolous:
Implement a blacklist for a public peergos server based on username. This stops those users reading or writing from this server, and also stops public links to files owned by them.

### DIFF
--- a/src/peergos/server/Start.java
+++ b/src/peergos/server/Start.java
@@ -1,13 +1,13 @@
 package peergos.server;
 
 import peergos.server.corenode.*;
+import peergos.server.mutable.*;
 import peergos.shared.*;
 import peergos.shared.corenode.*;
 import peergos.shared.crypto.asymmetric.*;
 import peergos.shared.crypto.asymmetric.curve25519.*;
 import peergos.shared.mutable.*;
 import peergos.shared.storage.*;
-import peergos.server.mutable.PinningMutablePointers;
 import peergos.server.fuse.*;
 import peergos.server.storage.*;
 import peergos.server.tests.*;
@@ -109,7 +109,9 @@ public class Start
 
             CoreNode core = HTTPCoreNode.getInstance(coreAddress);
             MutablePointers mutable = HttpMutablePointers.getInstance(coreAddress);
-            MutablePointers pinner = new PinningMutablePointers(mutable, dht);
+            String blacklistPath = "blacklist.txt";
+            PublicKeyBlackList blacklist = new UserBasedBlackList(Paths.get(blacklistPath), core, mutable, dht);
+            MutablePointers pinner = new BlockingMutablePointers(new PinningMutablePointers(mutable, dht), blacklist);
 
             InetSocketAddress httpsMessengerAddress = new InetSocketAddress(hostname, userAPIAddress.getPort());
             new UserService(httpsMessengerAddress, Logger.getLogger("IPFS"), dht, core, pinner, a);
--- a/src/peergos/server/Start.java
+++ b/src/peergos/server/Start.java
@@ -110,7 +110,7 @@ public class Start
             CoreNode core = HTTPCoreNode.getInstance(coreAddress);
             MutablePointers mutable = HttpMutablePointers.getInstance(coreAddress);
             String blacklistPath = "blacklist.txt";
-            PublicKeyBlackList blacklist = new UserBasedBlackList(Paths.get(blacklistPath), core, mutable, dht);
+            PublicKeyBlackList blacklist = new UserBasedBlacklist(Paths.get(blacklistPath), core, mutable, dht);
             MutablePointers pinner = new BlockingMutablePointers(new PinningMutablePointers(mutable, dht), blacklist);
 
             InetSocketAddress httpsMessengerAddress = new InetSocketAddress(hostname, userAPIAddress.getPort());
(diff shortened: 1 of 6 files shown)


src/peergos/server/mutable/PublicKeyBlackList.java
```diff
@@ -0,0 +1,8 @@
+package peergos.server.mutable;
+
+import peergos.shared.crypto.hash.*;
+
+public interface PublicKeyBlackList {
```
kevodwyer: Blacklist rather than BlackList?

src/peergos/server/mutable/UserBasedBlackList.java
```diff
@@ -0,0 +1,128 @@
...
+    private void updateBlackList() {
+        long modified = source.toFile().lastModified();
+        long now = System.currentTimeMillis();
+        if (modified != lastModified || (now - lastReloaded > RELOAD_PERIOD)) {
+            System.out.println("Updating blacklist...");
+            lastModified = modified;
+            lastReloaded = now;
+            Set<String> usernames = readUsernameFromFile();
+            Set<PublicKeyHash> updated = buildBlackList(usernames);
+            banned.clear();
```
kevodwyer: race condition on clear

src/peergos/server/mutable/UserBasedBlackList.java
```diff
@@ -0,0 +1,128 @@
...
+            for (PublicKeyHash hash : updated) {
+                banned.put(hash, true);
+            }
+        }
+    }
+
+    private Set<String> readUsernameFromFile() {
```
kevodwyer: Usernames with an s

src/peergos/server/Start.java
```diff
@@ -109,7 +109,9 @@ public static void startPeergos(Args a) {
 
             CoreNode core = HTTPCoreNode.getInstance(coreAddress);
             MutablePointers mutable = HttpMutablePointers.getInstance(coreAddress);
-            MutablePointers pinner = new PinningMutablePointers(mutable, dht);
+            String blacklistPath = "blacklist.txt";
```
cboddy: Can this be parameterized?

src/peergos/server/mutable/UserBasedBlacklist.java
```diff
@@ -0,0 +1,128 @@
...
+
+import java.io.*;
+import java.nio.file.*;
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.stream.*;
+
+public class UserBasedBlacklist implements PublicKeyBlackList {
+
+    private static final long RELOAD_PERIOD = 3_600_000;
```
cboddy: ``RELOAD_PERIOD_MS``?

src/peergos/server/mutable/UserBasedBlacklist.java
```diff
@@ -0,0 +1,128 @@
...
+    private final Path source;
+    private final ForkJoinPool pool = new ForkJoinPool(1);
+    private long lastModified, lastReloaded;
+
+    public UserBasedBlacklist(Path source, CoreNode core, MutablePointers mutable, ContentAddressedStorage dht) {
+        this.source = source;
+        this.core = core;
+        this.mutable = mutable;
+        this.dht = dht;
+        pool.submit(() -> {
```
cboddy: I'd suggest  making this more resilient  to failures:
```
while (true) {
                try {
                    updateBlackList();
                } catch (Throwable t) {
                    //log t
                }

                try {
                    Thread.sleep(5000);
                } catch (InterruptedException e) {}
            }
```

src/peergos/server/mutable/UserBasedBlacklist.java
```diff
@@ -0,0 +1,128 @@
...
+            banned.clear();
+            for (PublicKeyHash hash : updated) {
+                banned.put(hash, true);
+            }
+        }
+    }
+
+    private Set<String> readUsernamesFromFile() {
+        try {
+            Stream<String> lines = Files.lines(source);
```
cboddy: Could also do

```
return Files.lines(source).map(String::trim).collect(Collectors.toSet());
```

src/peergos/server/mutable/UserBasedBlacklist.java
cboddy: Also I think there's a race-condition here: if ``this.isAllowed`` is called before the pool executes ``updateBlacklist`` for the first time. We could add a synchronous  call of ``updateBlacklist`` to the constructor before  submitting the runnable to ``pool``?

src/peergos/server/Start.java
ianopolous: I think it's fine to require the blacklist.txt name no? They can always symlink if they really want it elsewhere.

src/peergos/server/mutable/UserBasedBlacklist.java
ianopolous: done

src/peergos/server/mutable/UserBasedBlacklist.java
ianopolous: Yeah this is not thread safe. E.g. I clear the map before updating it, without a lock. But I'm not stressed about exact guarantees, so long as it works 99% of the time.

src/peergos/server/mutable/UserBasedBlacklist.java
ianopolous: nice.

PR #269, opened 2017-11-11
restrict usernames further to lowercase and no .'s

ianopolous:


src/peergos/server/corenode/UsernameValidator.java
```diff
@@ -9,17 +11,18 @@
...
-     * no _ or . at the beginning
+     * no _- at the end
+     * allowed characters [a-z0-9_-]
+     * no __ or -- or _- or -_ inside
+     * no _- at the beginning
      * is 1-32 characters long
      * @param username
      * @return true iff username is a valid username.
      */
+    @JsMethod
```
cboddy: I  don't think  this can  have the ``JsMethod`` annotation unless you've added ``Pattern`` implementation  to GWT?

src/peergos/server/tests/CorenodeTests.java
```diff
@@ -108,16 +108,29 @@ public void writeThroughput() throws Exception {
 
     @Test
     public void isValidUsernameTest() {
-        List<String> areValid = Arrays.asList("chris", "super_califragilistic_ex", "z", "c.h_r.i.s");
+        List<String> areValid = Arrays.asList(
+                "chris",
```
cboddy: nit: 8 spaces instead of 4.

src/peergos/server/corenode/UsernameValidator.java
```diff
@@ -9,17 +11,18 @@
...
 
     /** Username rules:
-     * no _ or . at the end
-     * allowed characters [a-zA-Z0-9._]
-     * no __ or _. or ._ or .. inside
-     * no _ or . at the beginning
+     * no _- at the end
+     * allowed characters [a-z0-9_-]
+     * no __ or -- or _- or -_ inside
+     * no _- at the beginning
```
cboddy: So ``.`` is allowed at the start of a username? Just checking.

ianopolous: . isn't allowed anywhere

src/peergos/server/corenode/UsernameValidator.java
ianopolous: fixed. Also was in a package that gwt doesn't compile (server)

PR #273, opened 2017-11-19
Feat/authenticated writes

ianopolous:
This pr makes all writes authenticated. This means that the data being written is signed. 
so the api is now
`put(signing key hash, signature, data)`

This means the server can choose to reject writes based on their owner. Currently implemented is a key filter that allows all registered users. A future pr will include an arbitrary white-list of usernames. 

This involved some careful reordering of operations during sign up and password change. Also required was a defensive cbor technique to prevent out of memory errors when trying to parse random bytes as cbor (we should probably use these everywhere to prevent another form of DOSsing). 

The commit messages should be quite helpful for understanding. It all stems from the ContentAddressedStorage interface.

### DIFF
--- a/src/peergos/server/UserService.java
+++ b/src/peergos/server/UserService.java
@@ -157,7 +157,7 @@ public class UserService
 
         Function<HttpHandler, HttpHandler> wrap = h -> !isLocal ? new HSTSHandler(h) : h;
 
-        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht)));
+        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht, new RegisteredUserKeyFilter(coreNode, mutable, dht))));
         server.createContext(SIGNUP_URL, wrap.apply(new InverseProxyHandler("startDemo.peergos.net", isLocal)));
         server.createContext(ACTIVATION_URL, wrap.apply(new InverseProxyHandler("startDemo.peergos.net", isLocal)));
 
--- a/src/peergos/server/UserService.java
+++ b/src/peergos/server/UserService.java
@@ -157,7 +157,7 @@ public class UserService
 
         Function<HttpHandler, HttpHandler> wrap = h -> !isLocal ? new HSTSHandler(h) : h;
 
-        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht, new RegisteredUserKeyFilter(coreNode, mutable, dht))));
+        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht, new RegisteredUserKeyFilter(coreNode, mutable, dht)::isAllowed)));
         server.createContext(SIGNUP_URL, wrap.apply(new InverseProxyHandler("startDemo.peergos.net", isLocal)));
         server.createContext(ACTIVATION_URL, wrap.apply(new InverseProxyHandler("startDemo.peergos.net", isLocal)));
 
--- a/src/peergos/server/UserService.java
+++ b/src/peergos/server/UserService.java
@@ -157,7 +157,7 @@ public class UserService
 
         Function<HttpHandler, HttpHandler> wrap = h -> !isLocal ? new HSTSHandler(h) : h;
 
-        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht, mutable)));
+        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht)));
         server.createContext(SIGNUP_URL, wrap.apply(new InverseProxyHandler("startDemo.peergos.net", isLocal)));
         server.createContext(ACTIVATION_URL, wrap.apply(new InverseProxyHandler("startDemo.peergos.net", isLocal)));
 
--- a/src/peergos/server/UserService.java
+++ b/src/peergos/server/UserService.java
@@ -157,7 +157,7 @@ public class UserService
 
         Function<HttpHandler, HttpHandler> wrap = h -> !isLocal ? new HSTSHandler(h) : h;
 
-        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht)));
+        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht, mutable)));
         server.createContext(SIGNUP_URL, wrap.apply(new InverseProxyHandler("startDemo.peergos.net", isLocal)));
         server.createContext(ACTIVATION_URL, wrap.apply(new InverseProxyHandler("startDemo.peergos.net", isLocal)));
 
--- a/src/peergos/server/mutable/HttpMutablePointerServer.java
+++ b/src/peergos/server/mutable/HttpMutablePointerServer.java
@@ -89,7 +89,7 @@ public class HttpMutablePointerServer
 
         void getPointer(DataInputStream din, DataOutputStream dout) throws Exception
         {
-            PublicKeyHash encodedSharingKey = PublicKeyHash.fromCbor(CborObject.deserialize(new CborDecoder(din)));
+            PublicKeyHash encodedSharingKey = PublicKeyHash.fromCbor(CborObject.deserialize(new CborDecoder(din), 1024));
             byte[] metadataBlob = mutable.getPointer(encodedSharingKey).get().orElse(new byte[0]);
 
             dout.write(metadataBlob);
--- a/src/peergos/server/mutable/HttpMutablePointerServer.java
+++ b/src/peergos/server/mutable/HttpMutablePointerServer.java
@@ -89,7 +89,7 @@ public class HttpMutablePointerServer
 
         void getPointer(DataInputStream din, DataOutputStream dout) throws Exception
         {
-            PublicKeyHash encodedSharingKey = PublicKeyHash.fromCbor(CborObject.deserialize(new CborDecoder(din), 1024));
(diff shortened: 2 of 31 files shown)


src/peergos/server/UserService.java
```diff
@@ -157,7 +157,7 @@ public void configure(HttpsParameters params) {
 
         Function<HttpHandler, HttpHandler> wrap = h -> !isLocal ? new HSTSHandler(h) : h;
 
-        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht)));
+        server.createContext(DHT_URL, wrap.apply(new DHTHandler(dht, new RegisteredUserKeyFilter(coreNode, mutable, dht))));
```
cboddy: Shouldn't this have a unique URL-stem?

src/peergos/server/net/DHTHandler.java
```diff
@@ -1,34 +1,39 @@
...
         this.apiPrefix = apiPrefix;
     }
 
-    public DHTHandler(ContentAddressedStorage dht) throws IOException {
-        this(dht, "/api/v0/");
+    public DHTHandler(ContentAddressedStorage dht, KeyFilter keyFilter) throws IOException {
+        this(dht, keyFilter, "/api/v0/");
     }
 
     private Map<String, List<String>> parseQuery(String query) {
```
cboddy: Nit:  this function is fairly long, having sub functions for each case would improve readability.

src/peergos/server/net/DHTHandler.java
```diff
@@ -73,7 +81,50 @@ public void handle(HttpExchange httpExchange) throws IOException {
...
+                            // Check if this is possible to cause an OOM error before parsing it as cbor
+                            if (PublicSigningKey.maybeValidKey(data.get(0))) {
+                                PublicSigningKey candidateKey = PublicSigningKey.fromByteArray(data.get(0));
+                                PublicKeyHash calculatedHash = dht.hashKey(candidateKey);
+                                if (calculatedHash.equals(writerHash)) {
+                                    candidateKey.unsignMessage(ArrayOps.concat(signatures.get(0), data.get(0)));
+                                    return candidateKey;
+                                }
+                            }
+                        } catch (Throwable e) {
```
cboddy: Catch something more specific?

src/peergos/server/storage/KeyFilter.java
```diff
@@ -0,0 +1,8 @@
+package peergos.server.storage;
+
+import peergos.shared.crypto.hash.*;
+
+public interface KeyFilter {
```
cboddy: This could be replaced with ``Predicate<PublicKeyHash>>`` which could be clearer.

src/peergos/server/tests/MerkleBtree.java
```diff
@@ -14,20 +16,32 @@
...
     }
 
-    public CompletableFuture<MerkleBTree> createTree(PublicKeyHash user) throws IOException {
+    public CompletableFuture<MerkleBTree> createTree(SigningPrivateKeyAndPublicHash user) throws IOException {
         return createTree(user, RAMStorage.getSingleton());
     }
 
-    public PublicKeyHash createUser() {
-        return PublicKeyHash.NULL;
+    public SigningPrivateKeyAndPublicHash createUser() {
```
cboddy: What is going on here?

src/peergos/server/tests/MerkleBtree.java
```diff
@@ -209,7 +223,7 @@ public void storageSize() throws Exception {
         }
 
         int size = ((RAMStorage)tree.storage).size();
-        if (size != 30)
+        if (size != 31)
```
cboddy: ``31`` is a nice prime, perhaps add a comment about why it is the expected value of ``size``?

src/peergos/server/net/DHTHandler.java
ianopolous: I tried that and it resulted in 5 different exception types before I gave up. In this case I don't care what went wrong, just that something went wrong.

src/peergos/server/net/DHTHandler.java
ianopolous: Agreed.

src/peergos/server/tests/MerkleBtree.java
ianopolous: I added 1 because of the signing key now being stored in the dht.

src/peergos/server/tests/UserPublicKeyLinkTests.java
```diff
@@ -51,8 +54,14 @@ public void testSerialization(UserPublicKeyLink link) {
     public void createChain() throws Exception {
         SigningKeyPair oldUser = SigningKeyPair.random(new SafeRandom.Java(), new Ed25519.Java());
         SigningKeyPair newUser = SigningKeyPair.random(new SafeRandom.Java(), new Ed25519.Java());
-        PublicKeyHash oldHash = ipfs.putSigningKey(oldUser.publicSigningKey).get();
-        PublicKeyHash newHash = ipfs.putSigningKey(newUser.publicSigningKey).get();
+        PublicKeyHash oldHash = ipfs.putSigningKey(
```
cboddy: How about extracting this method?
```
public PublicKeyHash putSingingKey(user User) {
     return ipfs.putSigningKey(
                user.secretSigningKey.signOnly(user.publicSigningKey.serialize()),
                ipfs.hashKey(user.publicSigningKey),
                user.publicSigningKey).get();
}

```

src/peergos/server/storage/KeyFilter.java
ianopolous: done

src/peergos/shared/crypto/asymmetric/curve25519/Ed25519SecretKey.java
```diff
@@ -43,10 +44,16 @@ public CborObject toCbor() {
...
     }
 
+    @Override
     public byte[] signMessage(byte[] message) {
         return implementation.crypto_sign(message, secretKey);
     }
 
+    @Override
+    public byte[] signOnly(byte[] message) {
+        return Arrays.copyOfRange(implementation.crypto_sign(message, secretKey), 0, TweetNaCl.SIGNATURE_SIZE_BYTES);
```
cboddy: ``Arrays.copyOf(signMessage(message), 0,  TweetNaCl.SIGNATURE_SIZE_BYTES);``

cboddy: Also, how about ``signatureOnly``?

src/peergos/shared/crypto/asymmetric/curve25519/Ed25519SecretKey.java
```diff
@@ -43,10 +44,16 @@ public CborObject toCbor() {
         return new CborObject.CborList(Arrays.asList(new CborObject.CborLong(type().value), new CborObject.CborByteArray(secretKey)));
     }
 
+    @Override
```
cboddy: What about adding a doc-string here 
``@returns message with signature prepended``

src/peergos/shared/crypto/asymmetric/curve25519/Ed25519SecretKey.java
ianopolous: done

src/peergos/shared/crypto/asymmetric/curve25519/Ed25519SecretKey.java
ianopolous: there is one on the interface

src/peergos/server/tests/UserPublicKeyLinkTests.java
ianopolous: done

src/peergos/server/mutable/HttpMutablePointerServer.java
```diff
@@ -89,7 +89,7 @@ void setPointer(DataInputStream din, DataOutputStream dout) throws Exception
 
         void getPointer(DataInputStream din, DataOutputStream dout) throws Exception
         {
-            PublicKeyHash encodedSharingKey = PublicKeyHash.fromCbor(CborObject.deserialize(new CborDecoder(din)));
+            PublicKeyHash encodedSharingKey = PublicKeyHash.fromCbor(CborObject.deserialize(new CborDecoder(din), 1024));
```
cboddy: Ah, good of ``2^10``. How  about a static constant?

src/peergos/server/tests/CborObjects.java
```diff
@@ -15,6 +16,16 @@
         return res;
     }
 
+    @Test
+    public void dosCborObject() throws Throwable {
+        // make a header for a byte[] that is 2^50 long
+        byte[] raw = ArrayOps.hexToBytes("5b0004000000000000");
+        try {
+            CborObject.fromByteArray(raw);
+            throw new Throwable("Should have failed!");
```
cboddy: http://junit.sourceforge.net/javadoc/org/junit/Assert.html#fail()

src/peergos/shared/crypto/asymmetric/SecretSigningKey.java
```diff
@@ -10,8 +10,20 @@
...
     PublicSigningKey.Type type();
 
+    /**
+     *
+     * @param message
+     * @return The signature + message
+     */
     byte[] signMessage(byte[] message);
 
+    /**
```
cboddy: w00t for docstrings :)

src/peergos/server/mutable/HttpMutablePointerServer.java
ianopolous: done

PR #275, opened 2017-11-29
feat/user_file_pinner: Periodically  pin all user files for users in a whitelist.

cboddy:


src/peergos/server/Start.java
```diff
@@ -113,13 +114,18 @@ public static void startPeergos(Args a) {
             MutablePointers mutable = HttpMutablePointers.getInstance(coreAddress);
             String blacklistPath = "blacklist.txt";
             PublicKeyBlackList blacklist = new UserBasedBlacklist(Paths.get(blacklistPath), core, mutable, dht);
-            MutablePointers pinner = new BlockingMutablePointers(new PinningMutablePointers(mutable, dht), blacklist);
+            MutablePointers mutablePointers = new BlockingMutablePointers(new PinningMutablePointers(mutable, dht), blacklist);
+
+            Path userPath = Paths.get(a.getArg("whitelist_path", "user_whitelist.txt"));
```
ianopolous: we need to pick a convention for arg names. So far we are camelCase everywhere, but I think I prefer a - to be consistent with linux et al. Thoughts?

PR #276, opened 2017-12-02
fix bug with follow followed by unfollow :)

ianopolous:


src/peergos/shared/user/UserContext.java
```diff
@@ -1027,7 +1027,7 @@ private FollowRequest decodeFollowRequest(byte[] raw) throws IOException {
                             return root;
                         });
             }
-            throw new IllegalStateException("Metadata blob not Present downloading entry point!");
+            return CompletableFuture.completedFuture(root);//throw new IllegalStateException("Metadata blob not Present downloading entry point!");
```
ianopolous: can we remove the exception rather than comment it? @kevodwyer 
And maybe log it to the console too (specifically the owner)

PR #287, opened 2018-01-13
fix spacing in Cid.java

jxub:


ianopolous:
Ideally we would have a javafmt, but until then I'm not stressed about whitespace.

PR #307, opened 2018-02-06
Fix/parallel space calc

ianopolous:
Parallelise space recalculation

### DIFF
--- a/src/peergos/server/SpaceCheckingKeyFilter.java
+++ b/src/peergos/server/SpaceCheckingKeyFilter.java
@@ -9,6 +9,7 @@ import peergos.shared.merklebtree.*;
 import peergos.shared.mutable.*;
 import peergos.shared.storage.*;
 import peergos.shared.user.*;
+import peergos.shared.util.*;
 
 import java.time.*;
 import java.util.*;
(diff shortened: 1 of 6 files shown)


ianopolous:
Can confirm this reduces the restart time on the demo server to 58s.

ianopolous:
Ok this is good to go now.

src/peergos/server/SpaceCheckingKeyFilter.java
```diff
@@ -113,14 +114,30 @@ public SpaceCheckingKeyFilter(CoreNode core,
     private void loadAllOwners() {
         try {
             List<String> usernames = core.getUsernames("").get();
+            if (usernames.size() == 0)
+                return;
+            int threads = Math.min(usernames.size(), 1000);
```
cboddy: Why 1000 threads?

src/peergos/shared/util/Exceptions.java
```diff
@@ -0,0 +1,14 @@
+package peergos.shared.util;
+
+import java.util.concurrent.*;
+
+public class Exceptions {
```
cboddy: test?

ianopolous: yes I have

PR #315, opened 2018-03-17
Feat/champ

ianopolous:
Implement and switch to a CHAMP

### DIFF
--- a/src/peergos/shared/hamt/Champ.java
+++ b/src/peergos/shared/hamt/Champ.java
@@ -54,6 +54,11 @@ public class Champ implements Cborable {
     }
 
     private static final Champ EMPTY = new Champ(new BitSet(), new BitSet(), new KeyElement[0]);
+
+    public static Champ empty() {
+        return EMPTY;
+    }
+
     private final BitSet dataMap, nodeMap;
     private final KeyElement[] contents;
 
@@ -114,7 +119,7 @@ public class Champ implements Cborable {
     }
 
 
-    CompletableFuture<MaybeMultihash> get(ByteArrayWrapper key, int depth, ContentAddressedStorage storage) {
+    public CompletableFuture<MaybeMultihash> get(ByteArrayWrapper key, int depth, ContentAddressedStorage storage) {
         final int bitpos = mask(key, depth);
 
         if (dataMap.get(bitpos)) { // local value
@@ -135,7 +140,7 @@ public class Champ implements Cborable {
         return CompletableFuture.completedFuture(MaybeMultihash.empty());
     }
 
-    CompletableFuture<Pair<Champ, Multihash>> put(SigningPrivateKeyAndPublicHash writer,
+    public CompletableFuture<Pair<Champ, Multihash>> put(SigningPrivateKeyAndPublicHash writer,
                                                   ByteArrayWrapper key,
                                                   int depth,
                                                   MaybeMultihash expected,
@@ -325,25 +330,21 @@ public class Champ implements Cborable {
                     .thenCompose(newChild -> {
                         // Todo shortcut here if hash == existing
 
-                        switch (newChild.left.contents.length) {
-                            case 0: {
-                                throw new IllegalStateException("Sub-node must have at least one element.");
-                            }
-                            case 1: {
-                                if (this.keyCount() == 0 && this.nodeCount() == 1) {
-                                    // escalate (singleton or empty) result
-                                    return CompletableFuture.completedFuture(newChild);
-                                } else {
-                                    // inline value (move to front)
-                                    Champ champ = copyAndMigrateFromNodeToInline(bitpos, newChild.left);
-                                    return storage.put(writer, champ.serialize()).thenApply(h -> new Pair<>(champ, h));
-                                }
-                            }
-                            default: {
-                                // modify current node (set replacement node)
-                                Champ champ = copyAndSetNode(bitpos, newChild);
+                        if (newChild.left.contents.length == 0) {
+                            throw new IllegalStateException("Sub-node must have at least one element.");
+                        } else if (depth > 0 && newChild.left.contents.length == 1) {
+                            if (this.keyCount() == 0 && this.nodeCount() == 1) {
+                                // escalate (singleton or empty) result
+                                return CompletableFuture.completedFuture(newChild);
+                            } else {
+                                // inline value (move to front)
+                                Champ champ = copyAndMigrateFromNodeToInline(bitpos, newChild.left);
                                 return storage.put(writer, champ.serialize()).thenApply(h -> new Pair<>(champ, h));
                             }
+                        } else {
+                            // modify current node (set replacement node)
+                            Champ champ = copyAndSetNode(bitpos, newChild);
+                            return storage.put(writer, champ.serialize()).thenApply(h -> new Pair<>(champ, h));
                         }
                     });
         }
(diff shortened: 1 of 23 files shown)


ianopolous:
The mains reasons for using this are:
1) The structure is independent of insertion order
2) There are no rebalance/splitting operations like in a btree
3) It is tunable (via bit-width) to using less space in a tree, or less space churn during it's construction

(1) means that there is no information leakage via insertion order that could reveal information about the directory structure to an attacker who can see a snapshot of the network. Though clearly an attacker logging the state of the network constantly can deduce insertion order anyway.
(1) Also means that the structure plays well with CRDTs layered on top of it. i.e. if two instances apply two updates in a different order, they will result in the same tree root. 
(2) means that the pin update will always have a consistent fast performance ~10ms, without the pathological cases due to btree rebalancing/splitting which can take >30s for a single call

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,445 @@
...
+        return masknbit(key, depth, 6);
+    }
+
+    private static int mask8bit(ByteArrayWrapper key, int depth) {
+        return key.data[depth] & 0xff;
+    }
+
+    private static int masknbit(ByteArrayWrapper key, int depth, int nbits) {
+        int index = depth * nbits / 8;
+        int shift = depth * nbits % 8;
```
kevodwyer: make precedence explicit

ianopolous: done

ianopolous:
Of course, the other thing I want to use the champ for is the PKI, because it will makes the structure eventually consistent, regardless of the order of users updating their keys/expiry. The only synchonization required at that point is to ensure uniqueness of new usernames.

ianopolous:
These files show the space usage and bandwidth usage for various choices of parameter for two trees, one with 1000 mappings, and the other with 10,000. 
[better.champ.usage.txt](https://github.com/Peergos/Peergos/files/1857252/better.champ.usage.txt)
[better.champ.usage.10000.txt](https://github.com/Peergos/Peergos/files/1857293/better.champ.usage.10000.txt)

To me, these suggest we choose one of the following sets of parameters:
3-bit 6 collisions
4-bit 5 collisions

Thoughts?

ianopolous:
According to the original paper it is faster than Scalas and Clojures HAMTs.

ianopolous:
See https://blog.acolyer.org/2015/11/27/hamt/

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+        dst[setIndex] = new HashPrefixPayload(updated, null);
+
+        return new Champ(dataMap, nodeMap, dst);
+    }
+
+    private Champ insertIntoPrefix(final int index, final ByteArrayWrapper key, final MaybeMultihash val) {
+        final HashPrefixPayload[] src = this.contents;
+        final HashPrefixPayload[] result = new HashPrefixPayload[src.length];
+
+        System.arraycopy(src, 0, result, 0, src.length);
```
cboddy: Also would be clearer with ``Arrays.copyOf``

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+        KeyElement[] prefix = new KeyElement[src[index].mappings.length + 1];
+        System.arraycopy(src[index].mappings, 0, prefix, 0, src[index].mappings.length);
+        prefix[prefix.length - 1] = new KeyElement(key, val);
+        // ensure canonical structure
+        Arrays.sort(prefix, Comparator.comparing(m -> m.key));
```
cboddy: Nice

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+    private Champ copyAndMigrateFromInlineToNode(final int bitpos, final Pair<Champ, Multihash> node) {
+
+        final int oldIndex = index(dataMap, bitpos);
+        final int newIndex = this.contents.length - 1 - index(nodeMap, bitpos);
+
+        final HashPrefixPayload[] src = this.contents;
+        final HashPrefixPayload[] dst = new HashPrefixPayload[src.length];
+
+        // copy 'src' and remove 1 element at position oldIndex and insert 1 element at position newIndex
+        assert oldIndex <= newIndex;
```
cboddy: I'd make this a full invariant and raise a ``RuntimeException``  or similar, since the asserts are elided with the ``-ea`` javac flag.

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+        System.arraycopy(src, 0, result, 0, insertIndex);
+        System.arraycopy(src, insertIndex, result, insertIndex + 1, src.length - insertIndex);
+        result[insertIndex] = new HashPrefixPayload(new KeyElement[]{new KeyElement(key, val)}, null);
+
+        BitSet newDataMap = BitSet.valueOf(dataMap.toByteArray());
+        newDataMap.set(bitpos);
+        return new Champ(newDataMap, nodeMap, result);
+    }
+
+    private Champ copyAndMigrateFromInlineToNode(final int bitpos, final Pair<Champ, Multihash> node) {
```
cboddy: docstring?

src/peergos/shared/hamt/Champ.java
cboddy: Same in other places with ``assert`` is used.

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+
+        final int oldIndex = index(dataMap, bitpos);
+        final int newIndex = this.contents.length - 1 - index(nodeMap, bitpos);
```
cboddy: I'm a  bit lost; is this a length or a position in the array?

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+
+        final HashPrefixPayload[] src = this.contents;
+        final HashPrefixPayload[] dst = new HashPrefixPayload[src.length];
+
+        // copy 'src' and remove 1 element at position oldIndex and insert 1 element at position newIndex
+        assert oldIndex <= newIndex;
+        System.arraycopy(src, 0, dst, 0, oldIndex);
+        System.arraycopy(src, oldIndex + 1, dst, oldIndex, newIndex - oldIndex);
```
cboddy: Is it possible for ``newIndex`` to be smaller than ``oldIndex``? If so this will fail.

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    private Champ overwriteChildLink(final int bitpos, final Pair<Champ, Multihash> node) {
+
+        final int setIndex = this.contents.length - 1 - index(nodeMap, bitpos);
+
+        final HashPrefixPayload[] src = this.contents;
+        final HashPrefixPayload[] dst = new HashPrefixPayload[src.length];
+
+        System.arraycopy(src, 0, dst, 0, src.length);
```
cboddy: 323-326 is repeated quite a bit, could extract  into a utility method (or just ``dst = Arrays.copyOf(src)``)

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+                        return err;
+                    }
+
+                    if (this.keyCount() == maxCollisions + 1 && this.nodeCount() == 0) {
+                    /*
+						 * Create new node with remaining pairs. The new node
+						 * will a) either become the new root returned, or b)
+						 * unwrapped and inlined during returning.
+						 */
+                        final BitSet newDataMap = BitSet.valueOf((depth == 0) ? dataMap.toByteArray() : new byte[0]);
```
cboddy: should this be checking ``depth !=0``?

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+        System.arraycopy(src, 0, dst, 0, src.length);
+        dst[setIndex] = new HashPrefixPayload(null, MaybeMultihash.of(node.right));
+
+        return new Champ(dataMap, nodeMap, dst);
+    }
+
+    public CompletableFuture<Pair<Champ, Multihash>> remove(SigningPrivateKeyAndPublicHash writer,
+                                                            ByteArrayWrapper key,
+                                                            byte[] hash,
+                                                            int depth,
```
cboddy: docstring wherefor art though? Specifically I'd really like to now what depth is.

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+                        boolean lastInPrefix = mappings.length == 1;
+                        if (depth == 0 && lastInPrefix)
+                            newDataMap.clear(bitpos);
+                        else
+                            newDataMap.set(mask(hash, 0, bitWidth));
+
+                        HashPrefixPayload[] src = this.contents;
+                        HashPrefixPayload[] dst = new HashPrefixPayload[src.length - (lastInPrefix ? 1 : 0)];
+                        System.arraycopy(src, 0, dst, 0, dataIndex);
+                        System.arraycopy(src, dataIndex + 1, dst, dataIndex + (lastInPrefix ? 0 : 1), src.length - dataIndex - 1);
```
cboddy: The general pattern used in several places:
1. copy contents
2. do-something-with-node-at-some-index

Could this be refactored into something that can reduce the line  count?

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+    private Champ copyAndMigrateFromNodeToInline(final int bitpos, final Champ node) {
+
+        final int oldIndex = this.contents.length - 1 - index(nodeMap, bitpos);
+        final int newIndex = index(dataMap, bitpos);
+
+        final HashPrefixPayload[] src = this.contents;
+        final HashPrefixPayload[] dst = new HashPrefixPayload[src.length];
+
+        // copy src and remove element at position oldIndex and insert element at position newIndex
+        assert oldIndex >= newIndex;
```
cboddy:
```
if (oldIndex >= newIndex)
    throw new RuntimeException("useful message")
```

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+                                                        new CborObject.CborNull()
+                                        ))
+                                        .collect(Collectors.toList()))))
+                        .collect(Collectors.toList()))
+        ));
+    }
+
+    public static Champ fromCbor(Cborable cbor) {
+        if (! (cbor instanceof CborObject.CborList))
+            throw new IllegalStateException("Invalid cbor for CHAMP! " + cbor);
```
cboddy: Add check on list-length==3 here too?  It would be a better error  messsage.

src/peergos/shared/user/ImmutableTree.java
```diff
@@ -0,0 +1,36 @@
+package peergos.shared.user;
+
+import peergos.shared.crypto.*;
+import peergos.shared.io.ipfs.multihash.*;
+import peergos.shared.merklebtree.*;
+
+import java.io.*;
+import java.util.concurrent.*;
+
+public interface ImmutableTree {
```
cboddy: docstring?

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+    }
+
+    private static class HashPrefixPayload {
+        public final KeyElement[] mappings;
+        public final MaybeMultihash link;
+
+        public HashPrefixPayload(KeyElement[] mappings, MaybeMultihash link) {
+            this.mappings = mappings;
+            this.link = link;
+            if ((mappings == null) ^ (link != null))
```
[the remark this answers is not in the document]
ianopolous: done, though we still need the check

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+    private static int mask(byte[] hash, int depth, int nbits) {
+        int index = (depth * nbits) / 8;
+        int shift = (depth * nbits) % 8;
+        int lowBits = Math.min(nbits, 8 - shift);
+        int hiBits = nbits - lowBits;
+        return ((hash[index] >> shift) & ((1 << lowBits) - 1)) |
+                ((hash[index + 1] & ((1 << hiBits) - 1)) << lowBits);
+    }
+
+    private static int index(BitSet bitmap, int bitpos) {
```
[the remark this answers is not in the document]
ianopolous: done

src/peergos/shared/hamt/Champ.java
```diff
@@ -0,0 +1,524 @@
...
+                                                         ByteArrayWrapper key,
+                                                         byte[] hash,
+                                                         int depth,
+                                                         MaybeMultihash expected,
+                                                         MaybeMultihash value,
+                                                         int bitWidth,
+                                                         int maxCollisions,
+                                                         Function<ByteArrayWrapper, byte[]> hasher,
+                                                         ContentAddressedStorage storage,
+                                                         Multihash ourHash) {
```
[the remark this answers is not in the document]
ianopolous: done

src/peergos/shared/hamt/Champ.java
ianopolous: done

src/peergos/shared/hamt/Champ.java
ianopolous: position in the array, counting from the end

ianopolous: (As described in the champ paper)

src/peergos/shared/hamt/Champ.java
ianopolous: Nope, because the inline mappings are all before the child mappings

src/peergos/shared/hamt/Champ.java
ianopolous: done

src/peergos/shared/hamt/Champ.java
ianopolous: nope

src/peergos/shared/hamt/Champ.java
ianopolous: done

src/peergos/shared/user/ImmutableTree.java
ianopolous: done

src/peergos/shared/hamt/Champ.java
ianopolous: done

PR #321, opened 2018-06-03
Video thumbnail

kevodwyer:


build.xml
```diff
@@ -207,12 +207,11 @@ Class-Path: ${manifest_cp}
     <copy todir="war/WEB-INF/lib" file="${gwt.sdk}/gwt-servlet-deps.jar" />
     <!-- Add any additional server libs that need to be copied -->
     <copy todir="war/WEB-INF/lib" file="lib/jnr-fuse-0.3-all.jar" />
-    <copy todir="war/WEB-INF/lib" file="lib/orchid-1.0.0.jar" />
   </target>
 
   <target name="javac" depends="libs" description="Compile java source to bytecode">
     <mkdir dir="war/WEB-INF/classes"/>
-    <javac srcdir="src" includes="**" encoding="utf-8"
+    <javac srcdir="src" includes="peergos/client/**" encoding="utf-8"
```
ianopolous: surely we need peergos/shared as well here?

src/peergos/gwt/emu/java/nio/file/StandardOpenOption.java
```diff
@@ -0,0 +1,7 @@
+package java.nio.file;
+
+public enum StandardOpenOption implements OpenOption {
```
ianopolous: is there a way we can avoid all these superfluous classes? Like have an interface, with a java impl in server, and a native JS one?

src/peergos/shared/user/fs/FileTreeNode.java
```diff
@@ -27,7 +29,7 @@
 public class FileTreeNode {
 
     final static int THUMBNAIL_SIZE = 100;
-
+    private enum FileType { Image, Video, Unknown };
```
ianopolous: isn't this covered by the mimetype?

src/peergos/shared/user/fs/FileTreeNode.java
```diff
@@ -302,6 +304,7 @@ private SigningPrivateKeyAndPublicHash getSigner() {
         return ((DirAccess) pointer.fileAccess).getChildrenLocations(pointer.filePointer.baseKey);
     }
 
+    @JsMethod
```
ianopolous: what needs this in JS?

src/peergos/shared/user/fs/FileTreeNode.java
```diff
@@ -1003,11 +1006,31 @@ public static FileTreeNode createRoot(TrieNode root) {
         return new byte[0];
     }
 
+    public static byte[] generateVideoThumbnail(byte[] videoBlob) {
```
ianopolous: is it possible to do this without creating a random file somewhere? 
also why is .mp4 hard coded?

src/peergos/shared/user/fs/FileTreeNode.java
```diff
@@ -1019,6 +1042,18 @@ public static FileTreeNode createRoot(TrieNode root) {
                             fut.complete(generateThumbnail(bytes));
                         });
                     }
+                }else if(type == FileType.Video) {
+                    if (network.isJavascript()) {
```
ianopolous: This is where we could use the interface, say, VideoThumbnailGenerator and put an instance of it in UserContext

build.xml
kevodwyer: it works without it! I don't understand, In any case i will add shared

src/peergos/gwt/emu/java/nio/file/StandardOpenOption.java
kevodwyer: Used by shared (FileTreeNode) rather then server. So in short no.

src/peergos/shared/user/fs/FileTreeNode.java
kevodwyer: not quite as there are many different image/video mimetypes. Using a generic enum rather then using mimeType.startsWith() everywhere seemed like a win.

src/peergos/shared/user/fs/FileTreeNode.java
kevodwyer: my mistake, fixed (was used in former web-ui branch that is now redundant)

src/peergos/shared/user/fs/FileTreeNode.java
kevodwyer: javafx Media class can only read from a file. I wish it read from a memory stream. From testing it seemed the file extension was important.

src/peergos/gwt/emu/java/nio/file/StandardOpenOption.java
ianopolous: But they're new, so I think we can avoid it by moving the video thumbnail java stuff to a tpye in server.

src/peergos/shared/user/fs/FileTreeNode.java
ianopolous: But then adding new thumbnails e.g. text, pdf, will all be breaking changes.

src/peergos/shared/user/fs/FileTreeNode.java
ianopolous: what about .webm or other extensions?

src/peergos/shared/user/fs/FileTreeNode.java
kevodwyer: The established pattern is to call method on the NativeJSThumbnail class. Return type is string (so can do base64 in gtw translated java), this is in contrast the java method  return byte[]

src/peergos/shared/user/fs/FileTreeNode.java
kevodwyer: javafx currently only supports limited set of video formats

ianopolous:
Ok this looks good to go, but have you investigated why the tests are timing out?

ianopolous:
Looks like the java video thumbnail test doesn't terminate for some reason?

ianopolous:
Do we need a Platform.exit() now after the test suite?

PR #323, opened 2018-06-20
Ipfs based corenode

ianopolous:
This implements an ipfs based core node. 

This means moving the pki to a champ with a zero padded identity as hash, and storing the root of this champ in mutable pointers under a new pki key. This pki key is owned by the peergos user, and occupies a new position on the WriterData (in named keys). This means we can rotate the pki key independently of the peergos user key without MITM possibility. 

I will be doing more local testing, but otherwise this is good to go.


src/peergos/server/corenode/IpfsCoreNode.java
```diff
@@ -0,0 +1,228 @@
...
+import java.net.*;
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.function.*;
+import java.util.stream.*;
+
+public class IpfsCoreNode implements CoreNode {
+
+    public static final PublicSigningKey PEERGOS_IDENTITY_KEY = PublicSigningKey.fromString("ggFYIE7uD1ViM9KfiA1w69n774/jk6hERINN3xACPyabWiBp");
+    public static final PublicKeyHash PEERGOS_IDENTITY_KEY_HASH = PublicKeyHash.fromString("zdpuAvZynWLuyvovJwa34bj24M7cspt5M8seFfrFLrPWDGFDW");
```
cboddy: As discussed it would  be great if this could be overridden/be configurable.

src/peergos/server/corenode/IpfsCoreNode.java
```diff
@@ -0,0 +1,228 @@
...
+        this.currentRoot = MaybeMultihash.empty();
+        this.ipfs = ipfs;
+        this.signer = signer;
+        this.mutable = mutable;
+        this.update(currentRoot);
+    }
+
+    private synchronized void update(MaybeMultihash newRoot) {
+        Consumer<Triple<ByteArrayWrapper, MaybeMultihash, MaybeMultihash>> consumer =
+                triple -> {
```
cboddy: Can you add a comment explaining what this is supposed to do (or extract into a method with a docstring)?

src/peergos/server/corenode/IpfsCoreNode.java
```diff
@@ -0,0 +1,228 @@
...
+                        ContentAddressedStorage ipfs,
+                        MutablePointers mutable) {
+        this.currentRoot = MaybeMultihash.empty();
+        this.ipfs = ipfs;
+        this.signer = signer;
+        this.mutable = mutable;
+        this.update(currentRoot);
+    }
+
+    private synchronized void update(MaybeMultihash newRoot) {
```
cboddy: Please add a docstring.

src/peergos/server/corenode/IpfsCoreNode.java
```diff
@@ -0,0 +1,228 @@
...
+        try {
+            Champ.applyToDiff(currentRoot, newRoot, consumer, ipfs).get();
+            this.currentRoot = newRoot;
+        } catch (Exception e) {
+            throw new RuntimeException(e.getMessage(), e);
+        }
+    }
+
+    @Override
+    public CompletableFuture<Boolean> updateChain(String username, List<UserPublicKeyLink> updatedChain) {
```
cboddy: docstring plz.

PR #340, opened 2018-10-18
Give all users a list of storage nodes in their PKI data.

ianopolous:
This is signed along with the username claim,
and allows users to bootstrap from nothing but the pki.

N.B. This is a breaking change.


ianopolous:
moved to https://github.com/Peergos/Peergos/pull/345

PR #344, opened 2018-10-30
Implement a file based content addressed storage for testing

ianopolous:
Fix fuse test issue with daylight saving.

Reduce heap size for tests

### DIFF
--- a/src/peergos/server/storage/FileContentAddressedStorage.java
+++ b/src/peergos/server/storage/FileContentAddressedStorage.java
@@ -1,5 +1,6 @@
 package peergos.server.storage;
 
+import peergos.server.util.Logging;
 import peergos.shared.cbor.*;
 import peergos.shared.crypto.hash.*;
 import peergos.shared.io.ipfs.cid.*;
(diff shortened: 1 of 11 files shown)


ianopolous:
This should fix all the random test crashes, which were running out of memory.

src/peergos/server/storage/FileContentAddressedStorage.java
```diff
@@ -0,0 +1,263 @@
...
+            Cid cid = new Cid(CID_V1, isRaw ? Cid.Codec.Raw : Cid.Codec.DagCbor, rawhash);
+            Path filePath = getFilePath(cid);
+            Path target = root.resolve(filePath);
+            Path parent = target.getParent();
+            File parentDir = parent.toFile();
+            if (! parentDir.exists() && ! parentDir.mkdirs())
+                throw new IllegalStateException("Couldn't create directory: " + parent);
+            for (Path someParent = parent; !someParent.equals(root); someParent = someParent.getParent()) {
+                final File someParentFile = someParent.toFile();
+                if ("root".equals(System.getProperty("user.name")) || ! someParentFile.canWrite()) {
```
cboddy: 1. Why check if  the user is ``root``? 
2. This is not a robust way to do that.
3. Should probably add a note somewhere about this not  working on platforms other than linux.

PR #347, opened 2018-11-05
Add owner key to mutable pointer lookups.

ianopolous:
This enables the p2p version to proxy to the correct ipfs node, via a corenode lookup of the owner.


src/peergos/server/net/DHTHandler.java
```diff
@@ -36,7 +36,7 @@ public DHTHandler(ContentAddressedStorage dht, BiFunction<PublicKeyHash, Integer
         this(dht, keyFilter, "/api/v0/");
     }
 
-    private Map<String, List<String>> parseQuery(String query) {
+    public static Map<String, List<String>> parseQuery(String query) {
```
cboddy: Should this be  moved into a utils class?

Also can  we document that it won't handle escape chars?

src/peergos/shared/NetworkAccess.java
```diff
@@ -78,10 +78,17 @@ public NetworkAccess clear() {
...
-    public static CompletableFuture<NetworkAccess> build(HttpPoster poster, boolean isJavascript) {
+    public static CoreNode buildProxyingCorenode(HttpPoster poster, Multihash pkiServerNodeId) {
+        return new HTTPCoreNode(poster, pkiServerNodeId);
+    }
+
+    public static CoreNode buildDirectCorenode(HttpPoster poster) {
+        return new HTTPCoreNode(poster);
+    }
+
+    public static CompletableFuture<NetworkAccess> build(CoreNode coreNode, HttpPoster poster, boolean isJavascript) {
```
cboddy: Nice

PR #350, opened 2018-12-04
Enable mirroring corenode for all non pki nodes

ianopolous:


src/peergos/server/tests/MultiNodeNetworkTests.java
```diff
@@ -92,7 +94,7 @@ public static void init() throws Exception {
                     .with("social-sql-file", ":memory:");
             Main.PEERGOS.main(normalNode);
 
-
+            Thread.sleep(10_000);
```
kevodwyer: why the sleep() call

ianopolous: Only way I could get the ipfs swarm connect calls not to error all the time.

ianopolous: (the ipfs daemon, despite being able to answer id calls over http, can't correctly answer this call yet)

src/peergos/shared/mutable/HttpMutablePointers.java
```diff
@@ -18,14 +18,29 @@
 
     private static final boolean LOGGING = true;
     private final HttpPoster direct, p2p;
+    private final String directUrlPrefix;
```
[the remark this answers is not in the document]
ianopolous: Yep! That's done in a later pr.

src/peergos/server/corenode/MirrorCoreNode.java
```diff
@@ -33,6 +32,10 @@ public MirrorCoreNode(CoreNode writeTarget,
         this.ipfs = ipfs;
         this.mutable = mutable;
         this.pkiOwnerIdentity = pkiOwnerIdentity;
+    }
+
+    public void start() {
+        running = true;
```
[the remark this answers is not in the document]
ianopolous: This method is never called outside tests. If there are problems like you mention we will fix them.

PR #359, opened 2018-12-27
Implement an auth mechanism for retrieving the encrypted follow requests

ianopolous:
This guards against someone retrieving and storing all follow requests,
and then decrypting them later with a large quantum computer.


ianopolous:
N.B. If they already have a large quantum computer then this doesn't help. This means that this is at best a temporary measure until we move to post quantum asymmetric crypto.

src/peergos/shared/user/UserContext.java
```diff
@@ -943,13 +943,10 @@ public boolean isJavascript() {
...
 
-    /**
-     * Process any responses to our follow requests.
-     *
-     * @return initial follow requests
-     */
-    public CompletableFuture<List<FollowRequest>> processFollowRequests() {
-        return network.social.getFollowRequests(signer.publicKeyHash).thenCompose(reqs -> {
+    private CompletableFuture<List<FollowRequest>> getFollowRequests() {
+        byte[] time = new CborObject.CborLong(System.currentTimeMillis()).serialize();
```
kevodwyer: I think i'm ok with this. browser client time should be utc and server time is utc. We haven't changed impl of System in gwt emu so it should be default impl.

src/peergos/server/net/MutationHandler.java
```diff
@@ -0,0 +1,73 @@
...
+import peergos.server.util.*;
+import peergos.shared.crypto.hash.*;
+import peergos.shared.mutable.*;
+import peergos.shared.util.*;
+
+import java.io.*;
+import java.util.*;
+import java.util.logging.*;
+
+public class MutationHandler implements HttpHandler {
```
cboddy: class docstring plz.

src/peergos/server/net/SocialHandler.java
```diff
@@ -0,0 +1,84 @@
...
+import peergos.shared.crypto.hash.*;
+import peergos.shared.social.*;
+import peergos.shared.util.*;
+
+import java.io.*;
+import java.util.*;
+import java.util.function.*;
+import java.util.logging.*;
+
+public class SocialHandler implements HttpHandler {
```
cboddy: class docstring plz.

src/peergos/server/net/MutationHandler.java
```diff
@@ -0,0 +1,73 @@
...
+                    break;
+                case "getPointer":
+                    byte[] metadataBlob = mutable.getPointer(owner, writer).get().orElse(new byte[0]);
+                    dout.write(metadataBlob);
+                    break;
+                default:
+                    throw new IOException("Unknown method "+ method);
+            }
+
+            dout.flush();
```
cboddy: should the flush/close be in a finally block?

cboddy: also the flush call is redundant since it is called during ``FilterOutputstream::close``

src/peergos/server/net/SocialHandler.java
```diff
@@ -0,0 +1,84 @@
...
+        ByteArrayOutputStream bout = new ByteArrayOutputStream();
+        DataOutputStream dout = new DataOutputStream(bout);
+
+        String path = exchange.getRequestURI().getPath();
+        if (path.startsWith("/"))
+            path = path.substring(1);
+        String[] subComponents = path.substring(UserService.SOCIAL_URL.length()).split("/");
+        String method = subComponents[0];
+        Map<String, List<String>> params = HttpUtil.parseQuery(exchange.getRequestURI().getQuery());
+        Function<String, String> last = key -> params.get(key).get(params.get(key).size() - 1);
```
cboddy: why not get the first  element?

src/peergos/server/net/SocialHandler.java
cboddy: what is this used by?

src/peergos/server/net/MutationHandler.java
cboddy: what is this used by?

src/peergos/shared/social/HttpSocialNetwork.java
```diff
@@ -40,62 +40,37 @@ private static String getProxyUrlPrefix(Multihash targetId) {
...
-                } catch (IOException e) {
-                    throw new RuntimeException(e);
-                }
-            });
-        } catch (IOException ioe) {
-            LOG.log(Level.WARNING, ioe.getMessage(), ioe);
-            return null;
-        }
+        return poster.get(urlPrefix + "social/getFollowRequests?owner=" + encode(owner.toString())
+                + "&auth=" + ArrayOps.bytesToHex(signedTime)).thenApply(res -> {
```
cboddy: ``signedTime`` is represented as a byte-array that is then  hex-encoded? What is ``signedTime``? If it represents an actual time I can think of easier ways  to represent it here.

src/peergos/shared/social/HttpSocialNetwork.java
```diff
@@ -40,62 +40,37 @@ private static String getProxyUrlPrefix(Multihash targetId) {
...
-                    return CoreNodeUtils.deserializeByteArray(din);
-                } catch (IOException e) {
-                    throw new RuntimeException(e);
-                }
-            });
-        } catch (IOException ioe) {
-            LOG.log(Level.WARNING, ioe.getMessage(), ioe);
-            return null;
-        }
+        return poster.get(urlPrefix + "social/getFollowRequests?owner=" + encode(owner.toString())
```
cboddy: what happens if this throws an  exception?

ianopolous:
We remove a follow request immediately after processing it. So a reply attack would only see the pending follow requests. 

To do a replay attack, an attacker would first have to MITM either the TLS connection to the Peergos server if the user is logged in through a public server, or the ~TLS connection between the IPFS nodes if you're logging in through someone else's Peergos server, neither of which is in scope of the threat model.

src/peergos/server/net/MutationHandler.java
ianopolous: This is used by the peergos server (UserService)

src/peergos/server/net/SocialHandler.java
ianopolous: This is used by the Peergos server (UserService)

src/peergos/server/net/SocialHandler.java
ianopolous: Later duplicate parameters override earlier ones.

src/peergos/shared/social/HttpSocialNetwork.java
ianopolous: SignedTime is a cryptographic signature of the serialized time in millis. How would you represent it? Bear in mind that the source call we are using is getUTCMillis in Java and Javascript, and we don't have all the fancy java time classes in the browser.

src/peergos/server/net/MutationHandler.java
ianopolous: It's to a byte array output stream so it never throws

ianopolous: because it's a dataoutput stream (which doesn't buffer) close is also redundant, so I've removed both calls in the later pr.

src/peergos/shared/social/HttpSocialNetwork.java
ianopolous: Then you get an exceptionally completed future. (Any method that returns a future is expected not to throw otherwise you are mixing sync and async programming)

PR #357, opened 2018-12-24
Transactions

ianopolous:
Implement transactions around chunks. This means blocks can't be lost before the block result is pinned (once IPFS implements the api)

PR #361, opened 2018-12-29
Make all locations relative

ianopolous:
This makes all locations in Peergos relative. That means the owner and writer are no longer explicit in child links. Both are present in entry points and public links still. and the writer will be present on a child link to a file with a different writing key (or the reverse parent link).

This paves the way for granting write access. 

This is now the 5th pending PR. 
Review these before this, @cboddy, as this depends on them :
https://github.com/Peergos/Peergos/pull/350
https://github.com/Peergos/Peergos/pull/354
https://github.com/Peergos/Peergos/pull/357
https://github.com/Peergos/Peergos/pull/359

### DIFF
--- a/src/peergos/shared/NetworkAccess.java
+++ b/src/peergos/shared/NetworkAccess.java
@@ -211,19 +211,19 @@ public class NetworkAccess {
         }
     }
 
-    public CompletableFuture<List<RetrievedCapability>> retrieveAllMetadata(List<Triple<PublicKeyHash, PublicKeyHash, Capability>> links) {
+    public CompletableFuture<List<RetrievedCapability>> retrieveAllMetadata(List<AbsoluteCapability> links) {
         List<CompletableFuture<Optional<RetrievedCapability>>> all = links.stream()
                 .map(link -> {
-                    PublicKeyHash owner = link.left;
-                    PublicKeyHash writer = link.right.writer.orElse(link.middle);
-                    byte[] mapKey = link.right.getMapKey();
+                    PublicKeyHash owner = link.owner;
+                    PublicKeyHash writer = link.writer;
+                    byte[] mapKey = link.getMapKey();
                     return tree.get(owner, writer, mapKey)
                             .thenCompose(key -> {
                                 if (key.isPresent())
                                     return dhtClient.get(key.get())
                                             .thenApply(dataOpt ->  dataOpt
                                                     .map(cbor -> new RetrievedCapability(
-                                                            link.right,
+                                                            link,
                                                             CryptreeNode.fromCbor(cbor, key.get()))));
                                 LOG.severe("Couldn't download link at: " + new Location(owner, writer, mapKey));
                                 Optional<RetrievedCapability> result = Optional.empty();
@@ -249,13 +249,13 @@ public class NetworkAccess {
     public CompletableFuture<Optional<FileWrapper>> retrieveEntryPoint(EntryPoint e) {
         return downloadEntryPoint(e)
                 .thenApply(faOpt ->faOpt.map(fa -> new FileWrapper(new RetrievedCapability(e.pointer, fa), e.ownerName,
-                        e.owner, e.pointer.writer.get(), e.pointer.signer)))
+                        e.pointer.signer)))
                 .exceptionally(t -> Optional.empty());
     }
 
     private CompletableFuture<Optional<CryptreeNode>> downloadEntryPoint(EntryPoint entry) {
         // download the metadata blob for this entry point
-        return tree.get(entry.owner, entry.pointer.writer.get(), entry.pointer.getMapKey()).thenCompose(btreeValue -> {
+        return tree.get(entry.pointer.owner, entry.pointer.writer, entry.pointer.getMapKey()).thenCompose(btreeValue -> {
             if (btreeValue.isPresent())
                 return dhtClient.get(btreeValue.get())
                         .thenApply(value -> value.map(cbor -> CryptreeNode.fromCbor(cbor,  btreeValue.get())));
--- a/src/peergos/shared/NetworkAccess.java
+++ b/src/peergos/shared/NetworkAccess.java
@@ -211,19 +211,21 @@ public class NetworkAccess {
         }
     }
 
-    public CompletableFuture<List<RetrievedCapability>> retrieveAllMetadata(List<Capability> links) {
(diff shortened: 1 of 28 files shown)


src/peergos/shared/user/UserContext.java
```diff
@@ -1112,16 +1114,16 @@ private FollowRequest decodeFollowRequest(byte[] raw) throws IOException {
...
 
         // need to to retrieve all the entry points of our friends
         return Futures.reduceAll(notOurFileSystemEntries, ourRoot, (t, e) -> addEntryPoint(ourName, null, t, e, network, random, fragmenter), (a, b) -> a)
                 .exceptionally(Futures::logError);
     }
 
-    public static EntryPoint convert(String owner, CapabilityWithPath cap) {
-        return new EntryPoint(cap.cap, owner, Collections.emptySet(), Collections.emptySet());
+    public static EntryPoint convert(String ownerName, PublicKeyHash owner, CapabilityWithPath cap) {
+        return new EntryPoint(cap.cap, ownerName, owner, Collections.emptySet(), Collections.emptySet());
```
kevodwyer: The 2 empty collections doesn't sound right

src/peergos/shared/user/fs/DirAccess.java
```diff
@@ -227,58 +239,75 @@ private static PaddedCipherText encryptChildren(SymmetricKey baseKey, List<Capab
...
-                    return res.addChildAndCommit(modified.capability, ourPointer.baseKey, ourPointer, signer, network, random);
-                });
+    public CompletableFuture<DirAccess> updateChildLink(PublicKeyHash ourOwner,
+                                                        Capability ourPointer,
+                                                        RetrievedCapability original,
+                                                        RetrievedCapability modified,
+                                                        SigningPrivateKeyAndPublicHash signer,
+                                                        NetworkAccess network,
+                                                        SafeRandom random) {
+        return removeChild(original, ourOwner, ourPointer, signer, network)
```
kevodwyer: do we want an exceptionally clause in here?

src/peergos/shared/user/UserContext.java
ianopolous: I'm considering getting rid of readers and writers on entry point, because it is information we should calculate from the capability caches.

src/peergos/shared/user/EntryPoint.java
```diff
@@ -90,7 +96,7 @@ public boolean equals(Object o) {
         EntryPoint that = (EntryPoint) o;
 
         if (pointer != null ? !pointer.equals(that.pointer) : that.pointer != null) return false;
-        if (owner != null ? !owner.equals(that.owner) : that.owner != null) return false;
```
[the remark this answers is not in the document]
ianopolous: fixed

src/peergos/shared/user/fs/DirAccess.java
ianopolous: I think the callers of this should handle that.

src/peergos/shared/NetworkAccess.java
```diff
@@ -211,19 +211,21 @@ public static NetworkAccess build(ContentAddressedStorage dht,
         }
     }
 
-    public CompletableFuture<List<RetrievedCapability>> retrieveAllMetadata(List<Capability> links) {
+    public CompletableFuture<List<RetrievedCapability>> retrieveAllMetadata(List<Triple<PublicKeyHash, PublicKeyHash, Capability>> links) {
```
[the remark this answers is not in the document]
ianopolous: That's done. It took a lot longer than expected due to some awkward bug I introduced during it.

src/peergos/shared/user/fs/Capability.java
```diff
@@ -14,49 +14,68 @@
...
+        this.mapKey = mapKey;
+        this.baseKey = baseKey;
+        this.signer = signer;
+    }
+
+    public Capability(byte[] mapKey, SymmetricKey baseKey) {
+        this(Optional.empty(), mapKey, baseKey, Optional.empty());
+    }
+
+    public Capability(PublicKeyHash writer, byte[] mapKey, SymmetricKey baseKey, Optional<SecretSigningKey> signer) {
```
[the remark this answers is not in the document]
ianopolous: I removed 3 of them.

src/peergos/shared/user/UserContext.java
ianopolous: done

ianopolous:
Locations here means the cryptree locations (SymmetricLocationLinks). This makes many operation constant time rather than O(subtree size). E.g. changing your password should (but doesn't) update your entire filesystem to have the new owner key. That is not necessary at all with this pr.

ianopolous:
Basically EntryPoints and shared capabilities use absolute locations, and all the internal cryptree locations are relative.

src/peergos/shared/user/fs/Capability.java
```diff
@@ -14,49 +14,68 @@
...
-        this(new Location(owner, writer.publicKeyHash, mapKey), Optional.of(writer.secret), baseKey);
+    @JsMethod
+    public byte[] getMapKey() {
+        return Arrays.copyOf(mapKey, mapKey.length);
     }
 
-    public Location getLocation() {
-        return location;
+    public Location getLocation(PublicKeyHash owner, PublicKeyHash writer) {
+        return new Location(owner, this.writer.orElse(writer), mapKey);
```
cboddy: > this.writer.orElse(writer)

I've been thinking about this for a few minutes. This method seems like a bug waiting to happen. What if instead we it split into two explicit methods: one for the entry-point case and one for the other case (where this.writer is empty)?

src/peergos/shared/user/EntryPoint.java
```diff
@@ -92,26 +92,20 @@ public static EntryPoint fromCbor(Cborable cbor) {
...
     }
 
     @Override
     public int hashCode() {
-        int result = pointer != null ? pointer.hashCode() : 0;
-        result = 31 * result + (ownerName != null ? ownerName.hashCode() : 0);
-        result = 31 * result + (readers != null ? readers.hashCode() : 0);
-        result = 31 * result + (writers != null ? writers.hashCode() : 0);
-        return result;
+        return Objects.hash(pointer, ownerName, owner, readers, writers);
```
cboddy: neat.

src/peergos/shared/user/fs/Capability.java
ianopolous: What is the bug you are thinking of? This is only called from one place currently: in a directory when we are retrieving all the children. In that case we need to use the directory's writer key, unless the value here is present (the child is under a different writing key).

PR #362, opened 2019-01-02
Clean up the follow request code,

ianopolous:
so that someone who isn't crazy can read it. This also resulted in the nice utilities AsymmetricCipherText and PaddedAsymmetricCipherText. 

Remove the legacy serialization classes DataSink and DataSource which
were created before we had cbor and all its beauty.

This is another breaking change.


src/peergos/shared/social/BlindFollowRequest.java
```diff
@@ -0,0 +1,34 @@
...
+
+    public BlindFollowRequest(PublicBoxingKey dummySource, PaddedAsymmetricCipherText cipher) {
+        this.dummySource = dummySource;
+        this.cipher = cipher;
+    }
+
+    @Override
+    public CborObject toCbor() {
+        Map<String, CborObject> result = new TreeMap<>();
+        result.put("b", dummySource.toCbor());
```
kevodwyer: strings b,k is a bit unclear

ianopolous: I've renamed them

PR #364, opened 2019-01-04
Implement the write cryptree

ianopolous:
This implements the write cryptree. 

This is much simpler than the read cryptree. Every folder or file gets a new symmetric key, the write base key. Then if a child has a different signing keypair than its parent then it will include a SymmetricLinkToSigner from its write base key to the signing key pair. 

Child links now include a new symmetric link from the dir base write key to the child base write key.


ianopolous:
I'm not convinced about the type structure, especially around RelativeCapability, AbsoluteCapability and WriteableAbsoluteCapability. Any ideas to make it better?

PR #365, opened 2019-01-07
Public sharing

ianopolous:
This gives users an ability to publish capabilities publicly within peergos, and they then become accessible through any Peergos server under the url:
http(s)://$peergos-server/public/$path-to-file


ianopolous:
This will be used by the "peergos" user to do releases of Peergos itself

PR #372, opened 2019-01-18
write access for files

kevodwyer:


src/peergos/shared/user/fs/CapabilityStore.java
```diff
@@ -69,20 +69,20 @@ private static boolean isEditCapFile(String filename) {
 
     public static CompletableFuture<FileWrapper> addReadOnlySharingLinkTo(FileWrapper sharedDir, AbsoluteCapability capability, NetworkAccess network, SafeRandom random,
                                                                    Fragmenter fragmenter) {
-        return addSharingLinkTo(sharedDir, capability.readOnly(), network, random, fragmenter, CapabilityStore.READ_SHARING_FILE_PREFIX,
+        return addSharingLinkTo(sharedDir, capability, network, random, fragmenter, CapabilityStore.READ_SHARING_FILE_PREFIX,
```
ianopolous: any reason the .readOnly() was moved from here to the caller of this function? It's safer to keep here.

kevodwyer: no good reason. happy for you to revert back

PR #383, opened 2019-01-24
AsyncLock

ianopolous:
Factor out idea to implement a lock that is held across multiple asynchronous tasks into AsyncLock

### DIFF
--- a/src/peergos/shared/util/AsyncLock.java
+++ b/src/peergos/shared/util/AsyncLock.java
@@ -1,0 +1,40 @@
+package peergos.shared.util;
+
+import java.util.concurrent.*;
+import java.util.function.*;
+
+/** This class implements a lock that can be held across multiple dependent asynchronous tasks which return a new value
+ * for the guarded object or to get the value after any pending updaters have completed
+ *
+ * @param <T>
+ */
+public class AsyncLock<T> {
+
+    private CompletableFuture<T> queueHead;
+
+    public AsyncLock(CompletableFuture<T> initialValue) {
+        this.queueHead = initialValue;
+    }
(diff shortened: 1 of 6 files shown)


src/peergos/shared/util/AsyncLock.java
```diff
@@ -0,0 +1,40 @@
...
+     * @param updater
+     * @return A future completed with the result from updater, or exceptionally completed on error
+     */
+    public synchronized CompletableFuture<T> runWithLock(Function<T, CompletionStage<T>> updater) {
+        CompletableFuture<T> existing = queueHead;
+        CompletableFuture<T> newHead = new CompletableFuture<>();
+        this.queueHead = newHead;
+
+        CompletableFuture<T> result = new CompletableFuture<>();
+        existing.thenCompose(current -> updater.apply(current)
```
kevodwyer: I have spent the last 30 minutes staring at this :) 
If the queueHead keeps changing, then complete() will only ever be called once. Doesn't that defeat the purpose of the complete() method to detect conflicts?

ianopolous: The point here is to ensure that .thenX can only be called once. If we didn't do this, then two callers could call .thenX on the same CF, and then both callers would get the same value and proceed concurrently. If both of those were also modifying operations then one would fail with a CAS exception. We are enforcing single holder of this lock.

ianopolous: To further expand CFs and more specifically CompletionStages are designed to have an arbitrary tree of dependent operations. Here we are enforcing that tree to be a linear chain.

PR #388, opened 2019-01-31
User level transactions for file uploads

ianopolous:


src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -399,58 +405,44 @@ public boolean isDirty() {
...
-        return uploadFileSection(filename, fileData, false, startIndex, endIndex, baseKey,
-                overwriteExisting, network, random, monitor, fragmenter);
+    public CompletableFuture<FileWrapper> uploadOrOverwriteFile(String filename,
+                                                                AsyncReader fileData,
+                                                                long length,
+                                                                NetworkAccess network,
+                                                                SafeRandom random,
+                                                                ProgressConsumer<Long> monitor,
+                                                                Fragmenter fragmenter,
+                                                                List<Location> locations) {
```
cboddy: Is it necessary to pass in ``locations``? 

Where it  is being called in the PR  the argument  could be replaced with``this.generateChildLocationsFromSize(length, random)`` AFAICT.

src/peergos/server/tests/UserTests.java
```diff
@@ -415,44 +421,6 @@ public void concurrentWritesToFile() throws Exception {
         checkFileContents(all, file, context);
     }
 
-    @Test
```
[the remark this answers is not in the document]
ianopolous: It was testing a feature I removed - the auto renaming of files upon upload if a concurrent upload with the same name exists

src/peergos/shared/user/fs/FileWrapper.java
ianopolous: in the pr yes, but the pr doesn't actually do any transactions. The parameter needs to be there for transactions.

ianopolous: Not sure what you mean by shared storage, but probably yes.

ianopolous: The real problem is principle of least privilege, because that overloaded method must have higher privilege than it needs.

ianopolous: One solution to that would be to give the transaction folder a different signing key. That would almost minimise the excess privilege being passed around.

PR #400, opened 2019-02-27
Test/filesystem simulator

cboddy:
[edit] this is  now ready for review @ianopolous @kevodwyer @cesarosum.

Note: this is  the first pass that only does I/O. The next step is permissioning other users.

@ianopolous if you want to make changes and/or to merge before next week please do.

### DIFF
--- a/scripts/ensure-compile.sh
+++ b/scripts/ensure-compile.sh
@@ -10,9 +10,9 @@ echo "********************************"
 echo " Checking if project compiles.."
 echo "********************************"
 
-if ! ant compile;
-then 
-    echo "Project failed to compile"
-    exit 1
-fi
+#if ! ant compile;
+#then 
+#    echo "Project failed to compile"
+#    exit 1
+#fi
 
--- a/scripts/ensure-compile.sh
+++ b/scripts/ensure-compile.sh
@@ -10,9 +10,9 @@ echo "********************************"
 echo " Checking if project compiles.."
 echo "********************************"
 
-#if ! ant compile;
-#then 
-#    echo "Project failed to compile"
-#    exit 1
-#fi
+if ! ant compile;
+then 
+    echo "Project failed to compile"
+    exit 1
+fi
 
--- a/src/peergos/server/tests/simulation/AccessControl.java
+++ b/src/peergos/server/tests/simulation/AccessControl.java
@@ -13,6 +13,10 @@ public interface AccessControl {
 
     void remove(Path path, String reader, FileSystem.Permission permission);
 
+    default List<String> getWithAncestors(Path path, FileSystem.Permission permission)  {
+
+    }
+
 
     /**
      * Model owners, reader and writers
@@ -34,6 +38,7 @@ public interface AccessControl {
     }
 
     default boolean can(Path path, String user, FileSystem.Permission permission) {
+
         return getOwner(path).equals(user) || get(path, permission).contains(user);
     }
 
--- a/src/peergos/server/tests/simulation/AccessControl.java
+++ b/src/peergos/server/tests/simulation/AccessControl.java
@@ -13,10 +13,6 @@ public interface AccessControl {
 
     void remove(Path path, String reader, FileSystem.Permission permission);
 
-    default List<String> getWithAncestors(Path path, FileSystem.Permission permission)  {
-
-    }
-
 
     /**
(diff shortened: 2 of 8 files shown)


src/peergos/server/tests/simulation/AccessControl.java
```diff
@@ -0,0 +1,90 @@
+package peergos.server.tests.simulation;
+
+import java.nio.file.Path;
+import java.util.*;
+
+public interface AccessControl {
+
+    String getOwner(Path path);
+
+    List<String> get(Path path, FileSystem.Permission permission);
```
ianopolous: Any reason why this is a list instead of a set?

src/peergos/server/tests/simulation/FileSystem.java
```diff
@@ -0,0 +1,33 @@
+package peergos.server.tests.simulation;
```
ianopolous: good package name!

src/peergos/server/tests/simulation/AccessControl.java
```diff
@@ -0,0 +1,90 @@
...
+
+        @Override
+        public String getOwner(Path path) {
+            return owners.get(path);
+        }
+
+        @Override
+        public List<String> get(Path path, FileSystem.Permission permission) {
+            switch (permission) {
+                case READ:
```
ianopolous: One thing to be careful of here is that write access implies read access in our model.

src/peergos/server/tests/simulation/AccessControl.java
```diff
@@ -0,0 +1,90 @@
...
+        }
+    }
+
+    default boolean can(Path path, String user, FileSystem.Permission permission) {
+        return getOwner(path).equals(user) || get(path, permission).contains(user);
+    }
+
+    class MemoryImpl implements AccessControl {
+        private Map<Path, String> owners = new HashMap<>();
+        public AccessControlUnit readers = new AccessControlUnit();
```
ianopolous: Another thing we'll eventually is transitive read access (which we definitely want) and revocation. If A owns file X and A grants B R(X), and then b grants C R(X). Then A revokes R(X) from B, that also revokes it from anyone downstream from B, including C.

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    private byte[] getNextFileContents() {
+        byte[] bytes = new byte[getNextFileLength()];
+        random.nextBytes(bytes);
+        return bytes;
+    }
+
+    public Simulator(int opCount, int seed, double readProbability, double mkdirProbability, double writeProbability, long meanFileLength, FileSystem referenceFileSystem, FileSystem testFileSystem) {
+        if (readProbability + mkdirProbability > 1.0)
```
ianopolous: You could check the sum of all three is within epsilon of 1.0

scripts/ensure-compile.sh
```diff
@@ -10,9 +10,9 @@ echo "********************************"
 echo " Checking if project compiles.."
 echo "********************************"
 
-if ! ant compile;
-then 
-    echo "Project failed to compile"
-    exit 1
-fi
+#if ! ant compile;
```
ianopolous: want to commit this?

src/peergos/server/tests/simulation/AccessControl.java
cboddy: No, not fussed really.

src/peergos/server/tests/simulation/Simulator.java
cboddy: this will change as we add more ops - will leave for now

ianopolous: So it will still be incorrect then?

scripts/ensure-compile.sh
cboddy: reverted

src/peergos/server/tests/simulation/AccessControl.java
cboddy: That's captured now

src/peergos/server/tests/simulation/Simulator.java
ianopolous: All good

PR #401, opened 2019-03-03
Upgrade to ipfs v0.4.19 for all platforms

ianopolous:


src/peergos/server/storage/IpfsInstaller.java
```diff
@@ -2,23 +2,45 @@
...
 
 /**
  * A Utility for installing IPFS.
  */
 public class IpfsInstaller {
 
     public enum DownloadTarget {
-        LINUX_AMD64 ("https://github.com/cboddy/ipfses/blob/master/linux/ipfs?raw=true",
-                new Multihash(Multihash.Type.sha2_256, ArrayOps.hexToBytes("fd7c8d05b806c3e8c129f5938fa1bd39ba0659c7e57c06ec5f86029836570b22")));
+        DARWIN_386("https://github.com/peergos/ipfs-releases/blob/master/v0.4.19/darwin-386/ipfs?raw=true",
```
cboddy: Have you tested that these URLs are correct in any way?

ianopolous: Yes I ran the main method...

ianopolous: And they are standard github urls. the same as the one we were using for our build.

ianopolous: Also, all the unit tests.

PR #406, opened 2019-03-08
Fixing thumbnail generation and upload

cesarosum:
Compiles and runs. Tested with two photos, 1.3MB and 15MB respectively and runs locally.

### DIFF
--- a/src/peergos/shared/user/fs/FileWrapper.java
+++ b/src/peergos/shared/user/fs/FileWrapper.java
@@ -719,39 +719,27 @@ public class FileWrapper {
                                                     fileWriteKey);
 
                                         return addChildPointer(filename, fileWriteCap, network, random, 2)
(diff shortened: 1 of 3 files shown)


src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -694,29 +694,78 @@ public static int getNumberOfChunks(long size) {
...
+                                    });
+                        })
+            );
+
         });
+
     }
 
+    private CompletableFuture<FileWrapper> generateThumbnailAndUpdate(
+            FileWrapper hashPointer,
```
ianopolous: parent is probably a clearer name than hashPointer

cesarosum: Agreed 👍

cesarosum:
@ianopolous Updated as per your comment.

PR #409, opened 2019-03-15
Grand unification of directory and file cryptree

ianopolous:
This PR does a lot of things:
1) Unify the dir and file serialization so that the network can't tell the difference between a dir and a file
2) In the process hide a few more things like whether or not there is a changed signing key pair for a given file (which happens when granting write access). 
3) Pad the file data / links to children in a directory to a multiple of 4k before encryption. This further obscures the size of chunks, which could have been used to deduce if a user had uploaded a file of a specific size modulo 5 MiB. 
4) Remove the usage of Fragmenter which was always the simple splitting one. We no longer erasure code chunks, because that should be done at a lower level. 
5) Remove the distinct metadata key for directories. Now we just protect the metadata with the parent key. I can't think of any downsides to this..

The critical class is FragmentedPaddedCipherText

I'd still like to allow small files and directories to inline their data/children links with an identity multihash, but there were issues with that which need to be fixed first.


ianopolous:
I've just fixed the identity multihash issues and used it to inline directory children and small files (< 4KiB)

src/peergos/shared/io/ipfs/cid/Cid.java
```diff
@@ -49,35 +49,30 @@ public static Codec lookup(long c) {
...
-        System.arraycopy(hashBytes, 0, res, index, hashBytes.length);
-        return Arrays.copyOfRange(res, 0, index + hashBytes.length);
+        try {
+            ByteArrayOutputStream res = new ByteArrayOutputStream();
+            putUvarint(res, version);
+            putUvarint(res, codec.type);
+            super.serialize(res);
+            return res.toByteArray();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
kevodwyer: Is runtime exception the best idea in this situation?

src/peergos/shared/io/ipfs/multihash/Multihash.java
```diff
@@ -69,29 +75,46 @@ public int compareTo(Multihash that) {
...
-        System.arraycopy(hash, 0, res, 2, hash.length);
-        return res;
+        try {
+            ByteArrayOutputStream res = new ByteArrayOutputStream();
+            putUvarint(res, type.index);
+            putUvarint(res, hash.length);
+            res.write(hash);
+            return res.toByteArray();
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
kevodwyer: I think i answered my own question about exception thrown from child class Cid class before... It ain't going to happen.

src/peergos/shared/user/fs/cryptree/CryptreeNode.java
```diff
@@ -7,104 +7,785 @@
...
+                WritableAbsoluteCapability newChildCap = new WritableAbsoluteCapability(ourNewCap.owner,
+                        ourNewCap.writer, newChildMapKey, newChildReadKey, newChildWriteKey);
+                return rfp.fileAccess.copyTo(rfp.capability, newChildReadKey,
+                        ourNewCap, newEntryWriter, ourNewParentKey, newChildMapKey, network, random, hasher)
+                        .thenCompose(newChildFileAccess -> {
+                            return dirFuture.thenCompose(dirAccess ->
+                                    dirAccess.addChildAndCommit(ourNewCap.relativise(newChildCap), ourNewCap,
+                                            newEntryWriter, network, random, hasher));
+                        });
+            }, (a, b) -> a.thenCompose(x -> b)); // TODO Think about this combiner function
```
kevodwyer: TODO msg

src/peergos/shared/util/AsyncLock.java
```diff
@@ -38,6 +38,7 @@ public AsyncLock(CompletableFuture<T> initialValue) {
                     updater.get().thenCompose(processor)
                             .thenApply(res -> newHead.complete(res) && result.complete(res))
                             .exceptionally(e -> newHead.complete(current) && result.completeExceptionally(e));
+                    t.printStackTrace();
```
kevodwyer: are you happy to leave stacktrace call in here?

src/peergos/server/tests/UserTests.java
```diff
@@ -311,13 +310,54 @@ public void renameFile() throws Exception {
...
+
+        String dirname = "somedir";
+        userRoot.mkdir(dirname, network, false, crypto.random, crypto.hasher).join();
+
+        FileWrapper dir = context.getByPath("/" + username + "/" + dirname).get().get();
+        RetrievedCapability pointer = dir.getPointer();
+        SymmetricKey baseKey = pointer.capability.rBaseKey;
+        SymmetricKey parentKey = dir.getParentKey();
+        Assert.assertTrue("parent key different from base key", ! parentKey.equals(baseKey));
+        pointer.fileAccess.getDirectChildren(baseKey, network).join();
```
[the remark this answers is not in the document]
ianopolous: Nope, it will throw if the key is wrong because the decryption will fail

src/peergos/shared/NetworkAccess.java
```diff
@@ -329,13 +336,15 @@ public static NetworkAccess build(ContentAddressedStorage dht,
...
     }
 
-    // this one
     public CompletableFuture<Multihash> uploadChunk(CryptreeNode metadata,
                                                     PublicKeyHash owner,
                                                     byte[] mapKey,
                                                     SigningPrivateKeyAndPublicHash writer,
                                                     TransactionId tid) {
         try {
+            System.out.println("Uploading chunk: " + (metadata.isDirectory() ? "dir" : "file")
```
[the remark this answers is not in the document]
ianopolous: Yes, but I think it's useful to leave in for now.

src/peergos/shared/io/ipfs/cid/Cid.java
ianopolous: It never throws an IOException because it is a byte array output stream.

src/peergos/shared/io/ipfs/multihash/Multihash.java
ianopolous: yes

src/peergos/shared/util/AsyncLock.java
ianopolous: Yep this took me ages to find because there was otherwise no log of the error

src/peergos/shared/user/fs/cryptree/CryptreeNode.java
ianopolous: Yep that's been there for a while. It's still a todo :-)

src/peergos/shared/crypto/FragmentedPaddedCipherText.java
```diff
@@ -0,0 +1,136 @@
...
+
+        if (cipherText.length <= 4096 + TweetNaCl.SECRETBOX_OVERHEAD_BYTES) {
+            // use inline identity hash for small amount of data (small files or directories)
+            FragmentWithHash frag = new FragmentWithHash(new Fragment(cipherText), hasher.identityHash(cipherText, true));
+            return new Pair<>(new FragmentedPaddedCipherText(nonce, Collections.singletonList(frag.hash)), Collections.singletonList(frag));
+        }
+
+        byte[][] split = split(cipherText, maxFragmentSize);
+
+        List<FragmentWithHash> frags = Arrays.asList(split).stream()
```
cboddy: nit: ``Stream.of(spilt)``

ianopolous: Cheers. I've changed it to Arrays.stream() (in the second PR)

PR #410, opened 2019-03-18
Key owner proofs, inline public keys and generalize capability store

ianopolous:
This PR does the following
1) Changes references in WriterData to owned keys to require a proof that we actually do own the key (owned key signing the owner key). This prevents anyone from forging ownership of other's keys, which could cause issues with quota and maybe more. This means the structures are now fully self authenticating as well!

2) All public key hashes now use the identity hash. This reduces the number of network round trips dramatically, makes the pki totally self contained (until we move to post-quantum keys), and improves privacy (no network lookups of other people's keys from their hash)

3) Generalizes the capability store to be able to handle any size capability (this PR changes the serialized size of caps). This simplifies the store into a single append only file of caps, rather than a series of files. Position is remembered by the number of bytes read.

### DIFF
--- a/src/peergos/shared/crypto/OwnerProof.java
+++ b/src/peergos/shared/crypto/OwnerProof.java
@@ -1,0 +1,64 @@
+package peergos.shared.crypto;
+
+import peergos.shared.*;
+import peergos.shared.cbor.*;
+import peergos.shared.crypto.hash.*;
+import peergos.shared.storage.*;
+
+import java.util.*;
+import java.util.concurrent.*;
+
+public class OwnerProof implements Cborable {
+    public final PublicKeyHash ownedKey;
+    public final byte[] signedOwner;
+
+    public OwnerProof(PublicKeyHash ownedKey, byte[] signedOwner) {
+        this.ownedKey = ownedKey;
+        this.signedOwner = signedOwner;
+    }
+
+    public CompletableFuture<PublicKeyHash> getOwner(ContentAddressedStorage ipfs) {
+        return ipfs.getSigningKey(ownedKey)
+                .thenApply(signer -> signer
+                        .map(k -> PublicKeyHash.fromCbor(CborObject.fromByteArray(k.unsignMessage(signedOwner))))
+                        .orElseThrow(() -> new IllegalStateException("Couldn't retrieve owned key: " + ownedKey)));
+    }
+
+    public static OwnerProof build(SigningPrivateKeyAndPublicHash ownedKeypair, PublicKeyHash owner) {
+        return new OwnerProof(ownedKeypair.publicKeyHash, ownedKeypair.secret.signMessage(owner.serialize()));
+    }
+
+    @Override
+    public CborObject toCbor() {
+        Map<String, CborObject> result = new TreeMap<>();
+        result.put("o", new CborObject.CborMerkleLink(ownedKey));
+        result.put("p", new CborObject.CborByteArray(signedOwner));
+        return CborObject.CborMap.build(result);
+    }
+
+    public static OwnerProof fromCbor(Cborable cbor) {
+        if (! (cbor instanceof CborObject.CborMap))
+            throw new IllegalStateException("Incorrect cbor for OwnerProof: " + cbor);
+
+        CborObject.CborMap m = (CborObject.CborMap) cbor;
+        PublicKeyHash ownedKey = m.get("o", PublicKeyHash::fromCbor);
+        byte[] proof = m.get("p", c -> (CborObject.CborByteArray) c).value;
+        return new OwnerProof(ownedKey, proof);
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        OwnerProof that = (OwnerProof) o;
+        return Objects.equals(ownedKey, that.ownedKey) &&
+                Arrays.equals(signedOwner, that.signedOwner);
+    }
+
+    @Override
+    public int hashCode() {
+        int result = Objects.hash(ownedKey);
+        result = 31 * result + Arrays.hashCode(signedOwner);
+        return result;
+    }
+}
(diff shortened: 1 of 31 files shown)


src/peergos/shared/user/fs/CapabilitiesFromUser.java
```diff
@@ -27,16 +27,17 @@ public long getRecordsRead() {
     @Override
     public CborObject toCbor() {
         Map<String, CborObject> cbor = new TreeMap<>();
-        cbor.put("count", new CborObject.CborLong(recordsRead));
+        cbor.put("count", new CborObject.CborLong(bytesRead));
```
kevodwyer: rename count -> offset

src/peergos/shared/user/FriendSourcedTrieNode.java
```diff
@@ -75,15 +75,15 @@ public FriendSourcedTrieNode(Supplier<CompletableFuture<FileWrapper>> homeDirSup
                 .thenCompose(sharedDirOpt -> {
                     if (!sharedDirOpt.isPresent())
                         return CompletableFuture.completedFuture(true);
-                    return CapabilityStore.getReadOnlyCapabilityCount(sharedDirOpt.get(), network)
+                    return CapabilityStore.getReadOnlyCapabilityFileSize(sharedDirOpt.get(), network)
                             .thenCompose(count -> {
```
kevodwyer: rename count -> offset

src/peergos/shared/user/fs/AsyncReader.java
```diff
@@ -30,6 +37,45 @@
...
+        ByteArrayInputStream in = new ByteArrayInputStream(buf);
+        return readIntoArray(buf, prefix.length, (int) Math.min((long)(buf.length - prefix.length), maxBytesToRead))
+                .thenCompose(bytesRead -> {
+                    for (int localOffset = 0; localOffset < bytesRead;) {
+                        try {
+                            CborObject readObject = CborObject.read(in, bytesRead);
+                            accumulator.accept(fromCbor.apply(readObject));
+                            localOffset += readObject.toByteArray().length;
+                        } catch (RuntimeException e) {
+                            e.printStackTrace();
```
kevodwyer: is the printStackTrace needed?

src/peergos/shared/user/fs/CapabilityStore.java
```diff
@@ -25,110 +25,46 @@
...
                                                                       WritableAbsoluteCapability capability,
                                                                       NetworkAccess network,
                                                                       SafeRandom random,
                                                                       Hasher hasher) {
-        return addSharingLinkTo(sharedDir, capability, network, random, hasher, CapabilityStore.EDIT_SHARING_FILE_PREFIX,
-                CapabilityStore.EDIT_CAPABILITY_SIZE, CapabilityStore.SHARING_EDIT_FILE_MAX_SIZE);
+        return addSharingLinkTo(sharedDir, capability, network, random, hasher, CapabilityStore.EDIT_SHARING_FILE_NAME);
     }
 
     public static CompletableFuture<FileWrapper> addSharingLinkTo(FileWrapper sharedDir,
```
kevodwyer: make method private?

src/peergos/shared/user/fs/CapabilityStore.java
```diff
@@ -245,15 +165,9 @@ private static boolean isEditCapFile(String filename) {
                                                                                               Hasher hasher,
                                                                                               long capIndex,
```
kevodwyer: rename method to ..FromOffset and change references from Index -> Offset

src/peergos/shared/user/fs/CapabilityStore.java
```diff
@@ -280,115 +189,96 @@ private static boolean isEditCapFile(String filename) {
...
+                                                                                  String ownerName,
+                                                                                  PublicKeyHash owner,
+                                                                                  AsyncReader reader,
+                                                                                  long maxBytesToRead,
+                                                                                  NetworkAccess network) {
+        if (maxBytesToRead == 0)
+            return CompletableFuture.completedFuture(Collections.emptyList());
+
+        List<AbsoluteCapability> caps = new ArrayList<>();
+        return reader.parseStream(AbsoluteCapability::fromCbor, caps::add, maxBytesToRead)
```
kevodwyer: should we version Capability?

ianopolous: I think we're okay because all the components are themselves versioned, and any top level change we could add a version to the cbor map then, with an absent version being v0.

src/peergos/shared/user/FriendSourcedTrieNode.java
```diff
@@ -98,12 +98,12 @@ public FriendSourcedTrieNode(Supplier<CompletableFuture<FileWrapper>> homeDirSup
     private synchronized CompletableFuture<Boolean> addEditableCapabilities(Optional<FileWrapper> sharedDirOpt, NetworkAccess network) {
         return CapabilityStore.getEditableCapabilityFileSize(sharedDirOpt.get(), network)
                 .thenCompose(editCount -> {
-                    if (editCount == capCountEdit)
+                    if (editCount == byteOffsetWrite)
```
kevodwyer: tiny cleaup editCount -> editBytes

src/peergos/shared/crypto/OwnerProof.java
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public static OwnerProof build(SigningPrivateKeyAndPublicHash ownedKeypair, PublicKeyHash owner) {
+        return new OwnerProof(ownedKeypair.publicKeyHash, ownedKeypair.secret.signMessage(owner.serialize()));
+    }
+
+    @Override
+    public CborObject toCbor() {
+        Map<String, CborObject> result = new TreeMap<>();
+        result.put("o", new CborObject.CborMerkleLink(ownedKey));
```
cboddy: Would this work?
```
Map<String, ? extends Cborable> result = new HashMap<>();
        result.put("o", ownedKey);
        result.put("p", signedOwner);
        return CborObject.CborMap.build(result);
```

PR #411, opened 2019-03-20
Implement a signup filter / limit policy

ianopolous:
Use this to not download the world when calculating storage usage in space checking key filter


src/peergos/server/tests/UserTests.java
```diff
@@ -60,6 +60,7 @@ public static Args buildArgs() {
             return Args.parse(new String[]{
                     "-port", Integer.toString(port),
                     "-logToConsole", "true",
+                    "max-users", "10000",
```
kevodwyer: why so high?

PR #412, opened 2019-03-25
Move entry points from friends into a file

ianopolous:
This makes the static data constant size (only 1 entry point until we implement dual login)


ianopolous:
@kevodwyer Which comments? There is still static data, just much smaller and bounded.

kevodwyer:
in method UserContext.processFollowRequests()

PR #415, opened 2019-03-27
introduce WriteSynchronizer

kevodwyer:


src/peergos/shared/user/UserContext.java
```diff
@@ -349,17 +350,18 @@ public boolean isJavascript() {
...
         progressCallback.accept("Retrieving Friends");
-        return userData.getValue()
-                .thenCompose(wd -> createFileTree(entrie, username, network, crypto.random, crypto.hasher)
+        return writeSynchronizer.getCurrentWriterData(signer.publicKeyHash, signer.publicKeyHash,
+                wd -> createFileTree(entrie, username, network, crypto.random, crypto.hasher)
                         .thenCompose(root -> {
                             this.entrie = root;
                             return getByPath("/" + username + "/" + "shared")
-                                    .thenCompose(sharedOpt -> {
+                                    .thenApply(sharedOpt -> {
```
ianopolous: this should still be thenCompose I believe

src/peergos/shared/user/UserContext.java
```diff
@@ -349,17 +350,18 @@ public boolean isJavascript() {
...
                                         if (!sharedOpt.isPresent())
                                             throw new IllegalStateException("Couldn't find shared folder!");
                                         return buildSharedWithCache(sharedOpt.get(), this::getUserRoot);
-                                    }).thenApply(res -> this);
-                        }));
+                                    });
+                        }).thenApply(b -> wd)
```
ianopolous: don't need this thenApply

src/peergos/shared/user/UserContext.java
```diff
@@ -518,30 +520,33 @@ public boolean isJavascript() {
...
-                                        return network.coreNode.updateChain(username, claimChain)
-                                                .thenCompose(updatedChain -> {
-                                                    if (!updatedChain)
-                                                        throw new IllegalStateException("Couldn't register new public keys during password change!");
-
-                                                    return UserContext.ensureSignedUp(username, newPassword, network, crypto);
-                                                });
-                                    });
-                                }));
+                                ).thenCompose(newSignerHash ->
```
ianopolous: if only indentation has changed here can we revert it?

src/peergos/shared/user/UserContext.java
kevodwyer: done

src/peergos/shared/user/UserContext.java
kevodwyer: nope, the code structure has changed.

src/peergos/shared/user/UserContext.java
kevodwyer: i think it is necessary...

PR #417, opened 2019-03-28
Make sure we don't have two duplicate write synchronizers

ianopolous:
This was caused by the Network.clear() during signup.


kevodwyer:
That method on Network is called in a lot of places. Any further changes?

ianopolous:
I found it by changing AsyncLock to keep a list of all the futures ever in the queue. Then noticed when I hit the cas failure that the write synchronizer had different queues, and were in fact were different instances. After that it was easy to find with a breakpoint in the WriteSynchronizer constructor.

PR #432, opened 2019-04-05
Handle when a friend changes their identity key.

ianopolous:
This involves looking up their new key when logging in, to retrieve
their entry point correctly. AsyncLock needs to consider the owner as
well as the writer in the key.

Without this, when a friend changes their password you cannot log in!

### DIFF
--- a/src/peergos/shared/user/WriteSynchronizer.java
+++ b/src/peergos/shared/user/WriteSynchronizer.java
@@ -18,6 +18,7 @@ import java.util.function.Function;
 public class WriteSynchronizer {
     private final MutablePointers mutable;
     private final ContentAddressedStorage dht;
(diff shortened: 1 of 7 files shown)


src/peergos/shared/user/WriteSynchronizer.java
```diff
@@ -18,18 +18,18 @@
 public class WriteSynchronizer {
     private final MutablePointers mutable;
     private final ContentAddressedStorage dht;
-    private final Map<PublicKeyHash, AsyncLock<CommittedWriterData>> pending = new ConcurrentHashMap<>();
+    private final Map<Pair<PublicKeyHash, PublicKeyHash>, AsyncLock<CommittedWriterData>> pending = new ConcurrentHashMap<>();
```
kevodwyer: add a comment to explain the K,V elements

PR #436, opened 2019-04-09
Fix another case where a friend's identity has changed.

ianopolous:
Also handle a friend being offline.

I want to rework the whole follow request processing part, but this works for now

### DIFF
--- a/src/peergos/shared/user/RetrievedEntryPoint.java
+++ b/src/peergos/shared/user/RetrievedEntryPoint.java
@@ -1,0 +1,20 @@
+package peergos.shared.user;
+
+import peergos.shared.user.fs.*;
+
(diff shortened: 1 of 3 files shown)


src/peergos/shared/user/RetrievedEntryPoint.java
```diff
@@ -0,0 +1,20 @@
...
+    public final String path;
+    public final FileWrapper file;
+
+    public RetrievedEntryPoint(EntryPoint entry, String path, FileWrapper file) {
+        this.entry = entry;
+        this.path = path;
+        this.file = file;
+    }
+
+    public String getPath() {
```
kevodwyer: what's the point of the method if the member variable is public?

PR #438, opened 2019-04-10
handle simultaneous uploads

kevodwyer:


src/peergos/shared/user/UploadSynchronizer.java
```diff
@@ -0,0 +1,32 @@
...
+import peergos.shared.user.fs.FileWrapper;
+import peergos.shared.util.AsyncLock;
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
+
+public class UploadSynchronizer {
+
+    private final Map<String, AsyncLock<FileWrapper>> pending = new ConcurrentHashMap<>();
```
ianopolous: we can't have async locks outside the write synchronizer otherwise they will race

PR #440, opened 2019-04-11
make mime type detection available to JS

kevodwyer:


src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -1468,7 +1468,8 @@ public static FileWrapper createRoot(TrieNode root) {
         return fut;
     }
 
-    private CompletableFuture<String> getFileType(AsyncReader imageBlob) {
+    @JsMethod
```
ianopolous: I'm fine with this, but curious why you need it? We have the mime type on the FileProperties for files that have been uploaded. Can it be made static as well?

kevodwyer: oops, yes it is on the props. closing this PR as change is uneccesary

PR #444, opened 2019-04-18
Fix/composable mutations

ianopolous:
Allow complex mutations in the write synchronizer with multiple commit points.

This combined with refreshing the source dir prior to upload allows concurrent uploads to naturally queue, and succeed in turn.


src/peergos/shared/NetworkAccess.java
```diff
@@ -299,6 +327,25 @@ public static NetworkAccess build(ContentAddressedStorage dht,
         });
     }
 
+    public CompletableFuture<Optional<FileWrapper>> getFile(WriterData base,
+                                                            AbsoluteCapability cap,
+                                                            Optional<SigningPrivateKeyAndPublicHash> entryWriter,
+                                                            String ownerName) {
+        return getMetadata(base, cap)
+                .thenApply(faOpt -> faOpt.map(fa -> new FileWrapper(Optional.empty(),
```
kevodwyer: Since we no longer have FileAccess, DirAccess classes var references to faOpt, fa don't mean as much. But i have noticed fileAccess is used in many places

src/peergos/shared/user/MutableTree.java
```diff
@@ -17,26 +17,53 @@
      * @param sharingKey
      * @param mapKey
      * @param value
-     * @param tid
-     * @return the new root hash of the tree
+     * @return true when complete
```
kevodwyer: the return doc entry is incorrect

src/peergos/shared/user/Snapshot.java
```diff
@@ -0,0 +1,56 @@
+package peergos.shared.user;
+
+import peergos.shared.*;
+import peergos.shared.crypto.*;
+import peergos.shared.crypto.hash.*;
+
+import java.util.*;
+import java.util.concurrent.*;
+
+public class Snapshot {
```
kevodwyer: Not convinced about the class name. Needs to be more descriptive

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -91,6 +89,11 @@ public boolean equals(Object other) {
         return pointer.equals(((FileWrapper) other).getPointer());
     }
 
+    public CompletableFuture<FileWrapper> updated(WriterData wd, NetworkAccess network) {
```
kevodwyer: method name should include verb get

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -610,8 +655,7 @@ public static int getNumberOfChunks(long size) {
                 .collect(Collectors.toList());
     }
 
-    @JsMethod
-    public CompletableFuture<FileWrapper> uploadFileJS(String filename,
+    public CompletableFuture<FileWrapper> uploadFile(String filename,
```
kevodwyer: remember this change has a corresponding web-ui PR

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -1111,42 +1145,44 @@ public SigningPrivateKeyAndPublicHash signingPair() {
     }
 
     @JsMethod
-    public CompletableFuture<FileWrapper> copyTo(FileWrapper target, UserContext context) {
+    public CompletableFuture<Snapshot> copyTo(FileWrapper target, UserContext context) {
```
kevodwyer: this may impact web-ui code. Do we want to expose Snapshot in API?

src/peergos/shared/NetworkAccess.java
ianopolous: Semantically they are still as they were - file and directory accesses. They are functionally different. It's just that they have the same serialization now. I certainly still think of them as dir access and file access.

src/peergos/shared/user/MutableTree.java
ianopolous: fixed

src/peergos/shared/user/Snapshot.java
ianopolous: I've added some javadoc

src/peergos/shared/user/fs/FileWrapper.java
ianopolous: done

src/peergos/shared/user/fs/FileWrapper.java
ianopolous: I've restored the method name and annotation. No change required for UI.

src/peergos/shared/user/fs/FileWrapper.java
ianopolous: Good call. I've changed it to a boolean

ianopolous:
@cboddy and @cesarosum  No need to review this, but it should be of interest to you. The most important parts are Snapshot, WriteSynchronizer.applyComplexUpdate. The latter gives us a way to hold an async lock and commit partial results along the way. e.g. Uploading a file.

PR #462, opened 2019-04-26
Make Java also use caching mutable pointers, this speeds up reads by 6X!!

ianopolous:
The tests should also be much faster now.

### DIFF
--- a/src/peergos/shared/mutable/CachingPointers.java
+++ b/src/peergos/shared/mutable/CachingPointers.java
@@ -40,6 +40,13 @@ public class CachingPointers implements MutablePointers {
         synchronized (cache) {
             cache.remove(writer);
         }
-        return target.setPointer(ownerPublicKey, writer, writerSignedBtreeRootHash);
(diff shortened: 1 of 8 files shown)


src/peergos/shared/mutable/CachingPointers.java
```diff
@@ -40,6 +40,13 @@ public CachingPointers(MutablePointers target, int cacheTTL) {
         synchronized (cache) {
             cache.remove(writer);
         }
-        return target.setPointer(ownerPublicKey, writer, writerSignedBtreeRootHash);
+        return target.setPointer(ownerPublicKey, writer, writerSignedBtreeRootHash).thenApply(res -> {
+            if (res) {
+                synchronized (cache) {
+                    cache.put(writer, new Pair<>(Optional.of(writerSignedBtreeRootHash), System.currentTimeMillis()));
```
kevodwyer: Do we mind that the cache is unbounded as entries are never evicted?

PR #465, opened 2019-05-01
Clean up the api to remove duplicate methods and reduce number of args

ianopolous:


src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -95,7 +95,12 @@ public boolean equals(Object other) {
         return pointer.equals(((FileWrapper) other).getPointer());
     }
 
-    public CompletableFuture<FileWrapper> getUpdated(Snapshot version, NetworkAccess network) {
+    public CompletableFuture<FileWrapper> getUpdated(NetworkAccess network) {
```
kevodwyer: This overload feels a bit dangerous.

ianopolous: One is private, one is public.

ianopolous:
Yes there are a few js-methods with one less argument.

PR #496, opened 2019-06-03
Add extra salt to username when generating a user

ianopolous:


cesarosum:
Ouch that's a lot of files!

src/peergos/shared/user/ScryptGenerator.java
```diff
@@ -11,14 +11,22 @@
 
     @JsProperty
     public final int memoryCost, cpuCost, parallelism, outputBytes;
+    @JsProperty
+    public final String extraSalt;
 
-    public ScryptGenerator(int memoryCost, int cpuCost, int parallelism, int outputBytes) {
+    public ScryptGenerator(int memoryCost, int cpuCost, int parallelism, int outputBytes, String extraSalt) {
         if (memoryCost < MIN_MEMORY_COST)
```
cesarosum: Can we either add a comment or more information to the exception message about MIN_MEMORY_COST? Is it 17 bytes? Why that number?

src/peergos/shared/user/UserUtil.java
```diff
@@ -22,7 +22,7 @@
                                                                SecretGenerationAlgorithm algorithm) {
         if (password.equals(username))
             return Futures.errored(new IllegalStateException("Your password cannot be the same as your username!"));
-        CompletableFuture<byte[]> fut = hasher.hashToKeyBytes(username, password, algorithm);
+        CompletableFuture<byte[]> fut = hasher.hashToKeyBytes(username + algorithm.getExtraSalt(), password, algorithm);
```
cesarosum: For my own understanding, why add the salt to the username and not the password?

src/peergos/shared/user/RandomSecretType.java
```diff
@@ -24,4 +24,9 @@ static RandomSecretType fromCbor(CborObject cbor) {
     public Type getType() {
         return Type.Random;
     }
+
+    @Override
+    public String getExtraSalt() {
```
cesarosum: So `RandomSecretType` doesn't use an extra salt?

src/peergos/shared/crypto/password/PasswordProtected.java
```diff
@@ -13,7 +13,7 @@
 public class PasswordProtected {
 
     public static SecretGenerationAlgorithm getDefault() {
-        return new ScryptGenerator(ScryptGenerator.MIN_MEMORY_COST, 8, 1, 32);
+        return new ScryptGenerator(ScryptGenerator.MIN_MEMORY_COST, 8, 1, 32, "");
```
cesarosum: To be clear, we consider "no salt" to be the default behaviour?

src/peergos/shared/user/UserUtil.java
ianopolous: The username is the salt. We are just extending it.

src/peergos/shared/user/ScryptGenerator.java
ianopolous: 17 is the memory factor in the [scrypt](https://en.wikipedia.org/wiki/Scrypt) hashing function. It means the calculation will use 2^17 bytes of memory. This number was originally chosen to be sufficiently strong against brute force attacks, and to still take ~1s on a mobile.

src/peergos/shared/user/RandomSecretType.java
ianopolous: Indeed, it is already random, so we are no improving anything by adding salt.

src/peergos/shared/crypto/password/PasswordProtected.java
ianopolous: Current users have no extra salt, just the username as salt. New users after this pr, and existing users once they change their password will use a 32 random byte (converted to hex) salt.

src/peergos/shared/user/RandomSecretType.java
cesarosum: 👍

src/peergos/shared/crypto/password/PasswordProtected.java
cesarosum: Got it 👍

src/peergos/shared/user/UserUtil.java
cesarosum: Ah I see.

src/peergos/shared/user/ScryptGenerator.java
cesarosum: 👍

PR #501, opened 2019-06-14
Move to using blake2b to hash usernames in pki champ

ianopolous:
This moves from the identity hash to blake2b in the pki champ.
(Not a security issue, but will make it much easier to scale the pki, and shard it later on)

There is some temporary code in IpfsCoreNode to convert the existing champ to blake2b.

Currently the final commit of the conversion is commented until I'm confident it's correct.

### DIFF
--- a/src/peergos/shared/crypto/hash/Blake2b.java
+++ b/src/peergos/shared/crypto/hash/Blake2b.java
@@ -1,0 +1,2092 @@
+package peergos.shared.crypto.hash;
+
+// From https://github.com/alphazero/Blake2b
+/* !!! Doost !!! */
+
+/*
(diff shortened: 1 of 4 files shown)


src/peergos/shared/crypto/hash/Blake2b.java
```diff
@@ -0,0 +1,2092 @@
+package peergos.shared.crypto.hash;
+
+// From https://github.com/alphazero/Blake2b
+/* !!! Doost !!! */
```
cboddy: Doost?

ianopolous: Not sure what it means, but it was in the original copyright message

PR #513, opened 2019-07-06
new exists js function

kevodwyer:
required by upload folder web-ui change


ianopolous:
I would guess this to be saying if the instance itself exists. Maybe a name like hasChild or childExists?

PR #514, opened 2019-07-09
Implement friend annotations

ianopolous:
This allows a user to store (privately in their own space) annotations for each of their friends. 

Initially this is used for storing whether or not they have verified the keys of that user. 

It could be extended to add things like a local display name for a user.

### DIFF
--- a/src/peergos/shared/user/UserContext.java
+++ b/src/peergos/shared/user/UserContext.java
@@ -888,14 +888,18 @@ public class UserContext {
                     throw new RuntimeException(e);
                 }
             }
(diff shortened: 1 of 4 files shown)


src/peergos/shared/user/UserContext.java
```diff
@@ -284,18 +286,24 @@ public boolean isJavascript() {
                 .thenApply(x -> globalRoot);
     }
 
+    /**
+     *
+     * @param friendName
+     * @return a pair of the firends keys used to generate the fingerprint, and the resulting fingreprint
```
cesarosum: Nit: spelling

PR #515, opened 2019-07-10
Allow a native JS implementation of blake2b

ianopolous:


src/peergos/shared/fingerprint/FingerPrint.java
```diff
@@ -40,10 +40,11 @@ public String getBase64Thumbnail() {
     public static FingerPrint generate(String ourname,
                                        List<PublicKeyHash> ourIdentityKey,
                                        String friendsName,
-                                       List<PublicKeyHash> friendsIdentityKey) {
+                                       List<PublicKeyHash> friendsIdentityKey,
+                                       Hasher h) {
```
cesarosum: h is a little too terse. Perhaps `hasher`?

PR #534, opened 2019-07-12
Super fast seeking

ianopolous:
This PR implement a new scheme for files where the locations of subsequent chunks are deterministically calculated by:

next = hash(secret + current)
The secret is only stored in the initial chunk.

This means that we can seek to arbitrary offsets in a file with O(1) IO operations. Currently seeking in the web ui is super slow ~1s/10MiB. So seeking 1GiB in would take around 1.5 minutes!

There are several things to be careful of here:
1) to avoid downgrade attacks we have to make subsequent chunk calculations also use the hash calculation and not the pointer in the cryptree node
2) we need to rotate the stream secret when we rotate the base key (or the file key)
3) directories cannot use the same scheme, otherwise the size of the dir would be exposed to anyone with a read cap to any descendant of the dir

Obviously this is a fundamental change, so I want at least 2 reviews before merging.

I will also try blake2b instead of sha256, and use which ever is faster in the browser. 

There are a few minor changes to the JS api. FileWrapper.hasChild mainly I think.

### DIFF
--- a/src/peergos/server/tests/simulation/NativeFileSystemImpl.java
+++ b/src/peergos/server/tests/simulation/NativeFileSystemImpl.java
@@ -147,7 +147,10 @@ public class NativeFileSystemImpl implements FileSystem {
                 boolean isHidden = false;
                 String mimeType="NOT_SUPPORTED";
 
-                return new FileProperties(file.getName(), file.isDirectory(), mimeType, sizeHi, sizeLo, lastModified, isHidden, thumbnail);
+                //TODO make files use the new format with a stream secret
+                Optional<byte[]> streamSecret = file.isDirectory() ? Optional.empty() : Optional.empty();
+                return new FileProperties(file.getName(), file.isDirectory(), mimeType, sizeHi, sizeLo, lastModified,
+                        isHidden, thumbnail, streamSecret);
 
             }
 
--- a/src/peergos/shared/NetworkAccess.java
+++ b/src/peergos/shared/NetworkAccess.java
@@ -423,7 +423,13 @@ public class NetworkAccess {
         return Futures.combineAllInOrder(futures)
                 .thenApply(groups -> groups.stream()
                         .flatMap(g -> g.stream())
-                        .collect(Collectors.toList()));
+                        .collect(Collectors.toList()))
+                .thenApply(res -> {
+                    for (Fragment fragment : fragments) {
+                        FragmentedPaddedCipherText.ArrayCache.releaseArray(fragment.data);
+                    }
+                    return res;
+                });
     }
 
     public CompletableFuture<Multihash> uploadChunk(CryptreeNode metadata,
--- a/src/peergos/shared/NetworkAccess.java
+++ b/src/peergos/shared/NetworkAccess.java
@@ -104,7 +104,7 @@ public class NetworkAccess {
     }
 
     public static ContentAddressedStorage buildLocalDht(HttpPoster apiPoster, boolean isPeergosServer) {
-        return new CachingStorage(new ContentAddressedStorage.HTTP(apiPoster, isPeergosServer), 10_000, 50 * 1024);
+        return new CachingStorage(new ContentAddressedStorage.HTTP(apiPoster, isPeergosServer), 1_000, 50 * 1024);
     }
 
     @JsMethod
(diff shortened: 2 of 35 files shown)


ianopolous:
The other subtlety is that when rotating the stream secret during cleaning, all chunks but the first of a file will end up in a different location.

src/peergos/server/tests/simulation/NativeFileSystemImpl.java
```diff
@@ -147,7 +147,10 @@ public FileProperties fileProperties() {
                 boolean isHidden = false;
                 String mimeType="NOT_SUPPORTED";
 
-                return new FileProperties(file.getName(), file.isDirectory(), mimeType, sizeHi, sizeLo, lastModified, isHidden, thumbnail);
+                //TODO make files use the new format with a stream secret
+                Optional<byte[]> streamSecret = file.isDirectory() ? Optional.empty() : Optional.empty();
```
kevodwyer: I'm glad you added the comment

ianopolous:
I've confirmed that seeking 60mb in a video which was taking 6s is now taking about 3s (which is mostly the fixed overhead of a few seconds to get and decrypt the final chunk). I tried to test with a 4gb file, but we seem to have added a memory leak recently as both firefox and chrome crash now trying to upload it.

ianopolous:
After fixing the large file upload problem. I've confirmed that seeking even 1500 mb into a movie in the browser takes only 5s!

ianopolous:
Somehow the threadlocal works fine in the browser, but breaks in Java... If anyone can explain that to me please do.

src/peergos/shared/user/fs/LazyInputStreamCombiner.java
```diff
@@ -68,22 +68,22 @@ public LazyInputStreamCombiner(WriterData version,
...
-                .thenApply(p -> {
-                    updateState(0,globalIndex + Chunk.MAX_SIZE, p.left, p.right);
+                .thenApply(t -> {
+                    updateState(0,globalIndex + Chunk.MAX_SIZE, t.left, t.middle, t.right);
                     return true;
                 });
     }
 
-    private CompletableFuture<Pair<byte[], AbsoluteCapability>> getChunk(CryptreeNode access, int truncateTo) {
+    private CompletableFuture<Triple<byte[], byte[], AbsoluteCapability>> getChunk(CryptreeNode access, int truncateTo) {
```
cboddy: Make a data-class rather  than a 3-tuple? Clearer to read than the ``left``, ``middle``, ``right`` dotted over the place

src/peergos/shared/crypto/FragmentedPaddedCipherText.java
```diff
@@ -92,6 +92,12 @@ public static FragmentedPaddedCipherText fromCbor(Cborable cbor) {
                 .thenApply(fargs -> new CipherText(nonce, recombine(fargs)).decrypt(from, fromCbor));
     }
 
+    private static byte[][] generateCache() {
+        return new byte[40][Fragment.MAX_LENGTH];
```
cboddy: 40?

src/peergos/shared/crypto/FragmentedPaddedCipherText.java
```diff
@@ -92,6 +92,12 @@ public static FragmentedPaddedCipherText fromCbor(Cborable cbor) {
...
+    }
+
+    private static ThreadLocal<byte[][]> arrayCache = ThreadLocal.withInitial(FragmentedPaddedCipherText::generateCache);
```
cboddy: Is this ok with all the async going on?

src/peergos/shared/NetworkAccess.java
```diff
@@ -104,7 +104,7 @@ public static CoreNode buildDirectCorenode(HttpPoster poster) {
     }
 
     public static ContentAddressedStorage buildLocalDht(HttpPoster apiPoster, boolean isPeergosServer) {
-        return new CachingStorage(new ContentAddressedStorage.HTTP(apiPoster, isPeergosServer), 10_000, 50 * 1024);
+        return new CachingStorage(new ContentAddressedStorage.HTTP(apiPoster, isPeergosServer), 1_000, 50 * 1024);
```
cboddy: why?

PR #535, opened 2019-07-18
Add util methods to encrypt and decrypt urls in the web ui

ianopolous:


src/peergos/shared/user/UserContext.java
```diff
@@ -286,6 +286,35 @@ public boolean isJavascript() {
                 .thenApply(x -> globalRoot);
     }
 
+    @JsType
+    public static class EncryptedURL {
```
cboddy: Can we have a few unit tests for these, please? How about the classic (excuse the pseudocode):

```
x = EncryptedURL(a, b)
   assert x.decryptURLL(x.encryptURL(url)) == url
```

src/peergos/shared/user/UserContext.java
```diff
@@ -286,6 +286,35 @@ public boolean isJavascript() {
...
+        Base64.Encoder encoder = Base64.getEncoder();
+        return new EncryptedURL(encoder.encodeToString(nonce), encoder.encodeToString(cipherText));
+    }
+
+    @JsMethod
+    public String decryptURL(String cipherTextBase64, String nonceBase64) {
+        Base64.Decoder decoder = Base64.getDecoder();
+        byte[] nonce = decoder.decode(nonceBase64);
+        byte[] ciphertext = decoder.decode(cipherTextBase64);
+        return new String(rootKey.decrypt(ciphertext, nonce)).trim();
```
cboddy: Is trim safe here? What  if the thing ends with a space? If this is to stay then let's  blow up if someone calls ``encryptURL`` with something that ends with a space.

PR #543, opened 2019-08-17
Implement an option to mirror a user or a nodes data

ianopolous:
This can't include any pending follow requests, and the mutable pointers can't be retrieved except locally for now.

### DIFF
--- a/src/peergos/server/Main.java
+++ b/src/peergos/server/Main.java
@@ -362,15 +362,15 @@ public class Main {
             String path = mutablePointersSqlFile.equals(":memory:") ?
                     mutablePointersSqlFile :
                     a.fromPeergosDir("mutable-pointers-file").toString();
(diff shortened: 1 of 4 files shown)


src/peergos/server/Main.java
```diff
@@ -359,7 +362,8 @@ public static void startPeergos(Args a) {
             String path = mutablePointersSqlFile.equals(":memory:") ?
                     mutablePointersSqlFile :
                     a.fromPeergosDir("mutable-pointers-file").toString();
-            MutablePointers sqlMutable = UserRepository.buildSqlLite(path, localDht);
+            JdbcIpnsAndSocial mutableSql = new JdbcIpnsAndSocial(JdbcIpnsAndSocial.buildSqlLite(path));
```
kevodwyer: mutableSql and then sqlMutable, confusing!

ianopolous: I've renamed them

PR #544, opened 2019-08-21
[cli] get, put and ls working

cboddy:
This is a PoC for a CLI for Peergos that implements:
* logging in
* listing files
* getting files (download)
* putting files (upload)

This uses the [jline3](https://github.com/jline/jline3) library, (albeit in a minimal way so far). There are some comments for  the other intended commands and it should be easy to integrate the dynamic autocomplete more with the filesystem commands.
 
I'd recommend building off  this  branch and trying it in a shell to get a feel. Since intelliJ's in-built shell doesn't play nicely use:

```
ant && java -cp lib/*:PeergosServer.jar peergos.server.cli.CLI
```


ianopolous:
Awesome! Looks great! Super cool! Some minor nits:
* We don't want to depend on any "test" classes, so maybe move them to a non test package. 
* Can we use * imports rather than individual classes? It makes diffs much smaller and easier to read now and in the future.

ianopolous:
fyi, javac is failing in teamcity with the last commit

cboddy:
> fyi, javac is failing in teamcity with the last commit

should be fixed in later commit.

cboddy:
> Awesome! Looks great! Super cool! Some minor nits:
> 
>     * We don't want to depend on any "test" classes, so maybe move them to a non test package.
> 
>     * Can we use * imports rather than individual classes? It makes diffs much smaller and easier to read now and in the future.

Sure will do, fleshing out a  few more methods in the mean time.

src/peergos/server/cli/CLI.java
```diff
@@ -287,6 +289,12 @@ public String exit(ParsedCommand cmd) {
 
     }
 
+    public String space(ParsedCommand cmd) {
+        UserContext uc = cliContext.userContext;
+        Long spaceUsed = uc.getTotalSpaceUsed(uc.signer.publicKeyHash, uc.signer.publicKeyHash).join();
```
ianopolous: can do this in constant time with context.getSpaceUsage()

src/peergos/server/cli/CLI.java
```diff
@@ -45,7 +45,7 @@
         get_follow_requests("Show the users that have sent you a follow request"),
         follow("Send a follow-request to another user.", "follow username-to-follow"),
         passwd("Update your password"),
-        share("","");
+        share_read("share_read path <user>","Grant read access for a file to another user.");
```
ianopolous: nit: file or folder

cboddy:
> > Awesome! Looks great! Super cool! Some minor nits:
> > ```
> > * We don't want to depend on any "test" classes, so maybe move them to a non test package.
> > 
> > * Can we use * imports rather than individual classes? It makes diffs much smaller and easier to read now and in the future.
> > ```
> 
> Sure will do, fleshing out a few more commands in the mean time.

Done

src/peergos/server/cli/CLI.java
cboddy: Currently it is just files - will  update to include  folders in different PR.

ianopolous:
This is SO COOL! So much fun!

Special request: how about a "friends" command which list your friends (intersection of following and followers)? and maybe followers, and following too?

scripts/ensure-compile.sh
```diff
@@ -6,13 +6,13 @@
...
-echo " Checking if project compiles.."
-echo "********************************"
-
-if ! ant compile;
-then 
-    echo "Project failed to compile"
-    exit 1
-fi
-
+#echo "********************************"
```
ianopolous: do you want to commit this?

ianopolous:
Could we move the things that are testing only like Simulation back to a testing package?

ianopolous:
Super bonus marks if you can figure out how to do a progress bar for downloads/uploads (the api is there in peergos).

ianopolous:
Got this typing "?" hoping to get help:
java.lang.IllegalArgumentException: No enum constant peergos.server.cli.Command.?
Maybe catch that and say unknown command rather than crash and exit?

ianopolous:
More bonus marks for displaying the login messages (like we do in the webui)

ianopolous:
Can we hide hidden files from ls? At the moment hidden is more "system", and people shouldn't be messing with those files.

scripts/ensure-compile.sh
cboddy: No - reverted now.

cboddy:
> This is SO COOL! So much fun!
> 
> Special request: how about a "friends" command which list your friends (intersection of following and followers)? and maybe followers, and following too?


> More bonus marks for displaying the login messages (like we do in the webui)

Sure - a static message of an MotD? Can you point me at what you mean?

cboddy:
> Could we move the things that are testing only like Simulation back to a testing package?

Sure - if you tell me which classes will do?

cboddy:
> Super bonus marks if you can figure out how to do a progress bar for downloads/uploads (the api is there in peergos).

Yes indeed, I wanted this too - future work. given my holiday is almost up I'd suggest doing this and the "friends" and "?" as commands in a  separate PR.

cboddy:
> Got this typing "?" hoping to get help:
> java.lang.IllegalArgumentException: No enum constant peergos.server.cli.Command.?
> Maybe catch that and say unknown command rather than crash and exit?


Re ``?`` as a command - yeah I realised  using the enum name would exclude that too late but we can fix it. Both now done.

ianopolous:
> > This is SO COOL! So much fun!
> > Special request: how about a "friends" command which list your friends (intersection of following and followers)? and maybe followers, and following too?
> 
> > More bonus marks for displaying the login messages (like we do in the webui)
> 
> Sure - a static message of an MotD? Can you point me at what you mean?

You can pass a Consumer<String> to the signin and signup methods that prints the string.

ianopolous:
> > Could we move the things that are testing only like Simulation back to a testing package?
> 
> Sure - if you tell me which classes will do?

The classes that are only used by the fuzzer should be in a test package

ianopolous:
> Can we hide hidden files from ls? At the moment hidden is more "system", and people shouldn't be messing with those files.
> 
> Hmm, we can do this. Technically `/<user>/shared` is a hidden folder - do we want to add that to a static white-list of hidden files to be displayed?

All hidden things please, based on the hidden property. There are files and folders that users shouldn't mess with.

src/peergos/server/cli/CLI.java
ianopolous: Aren't folders exactly the same api? i.e. zero extra work?

cboddy:
> > Can we hide hidden files from ls? At the moment hidden is more "system", and people shouldn't be messing with those files.
> > Hmm, we can do this. Technically `/<user>/shared` is a hidden folder - do we want to add that to a static white-list of hidden files to be displayed?
> 
> All hidden things please, based on the hidden property. There are files and folders that users shouldn't mess with.

So, this is super easy to do.  It would be nice though if there was a way to see what had been shared with another user via the CLI (which ``/shared/<user?/`` provided). Is there another mechanism we can use?

ianopolous:
> > > Can we hide hidden files from ls? At the moment hidden is more "system", and people shouldn't be messing with those files.
> > > Hmm, we can do this. Technically `/<user>/shared` is a hidden folder - do we want to add that to a static white-list of hidden files to be displayed?
> > 
> > 
> > All hidden things please, based on the hidden property. There are files and folders that users shouldn't mess with.
> 
> So, this is super easy to do. It would be nice though if there was a way to see what had been shared with another user via the CLI (which `/shared/<user?/` provided). Is there another mechanism we can use?

UserContext.sharedWith(file/dir) is what you're looking for. Will tell you who a file has been shared with.

cboddy:
> > > This is SO COOL! So much fun!
> > > Special request: how about a "friends" command which list your friends (intersection of following and followers)? and maybe followers, and following too?
> > 
> > 
> > > More bonus marks for displaying the login messages (like we do in the webui)
> > 
> > 
> > Sure - a static message of an MotD? Can you point me at what you mean?
> 
> You can pass a Consumer to the signin and signup methods that prints the string.
> https://github.com/Peergos/Peergos/blob/master/src/peergos/shared/user/UserContext.java#L138
> and
> https://github.com/Peergos/Peergos/blob/master/src/peergos/shared/user/UserContext.java#L203

Ah nice, done.

cboddy:
> > > > Can we hide hidden files from ls? At the moment hidden is more "system", and people shouldn't be messing with those files.
> > > > Hmm, we can do this. Technically `/<user>/shared` is a hidden folder - do we want to add that to a static white-list of hidden files to be displayed?
> > > 
> > > 
> > > All hidden things please, based on the hidden property. There are files and folders that users shouldn't mess with.
> > 
> > 
> > So, this is super easy to do. It would be nice though if there was a way to see what had been shared with another user via the CLI (which `/shared/<user?/` provided). Is there another mechanism we can use?
> 
> UserContext.sharedWith(file/dir) is what you're looking for. Will tell you who a file has been shared with.

OK, that makes sense.  That's done - will leave showing the shared-with files as a separate command/PR.

src/peergos/server/cli/CLI.java
cboddy: Yes but at the moment everything in the CLI works for files and not folders which is a nice consistency to have.  When we update get/put etc. for folders we can remove do the same here.

ianopolous:
I think it's just moving the test classes, then this is good to go. (And awesome)

ianopolous:
Minor request, could ls -la print the sizes, and modified etc? Feel free to postpone to another pr.

ianopolous:
There's a new conflict

cboddy:
> There's a new conflict

trivially rebased onto master.

PR #545, opened 2019-08-21
First attempt at a transaction aware GC for ipfs

ianopolous:
This implements a transaction aware GC. 

We periodically queue a GC, then wait for all open transactions to complete or time out, before actually doing the gc, and forcing subsequent transaction opens to wait until the gc is completed. 

This can't be called remotely, despite being in ContentAddressedStorage. I'm open to any nicer ways you can think of to implement this without adding a method to CAS.

### DIFF
--- a/src/peergos/server/storage/GarbageCollector.java
+++ b/src/peergos/server/storage/GarbageCollector.java
@@ -17,6 +17,7 @@ public class GarbageCollector implements ContentAddressedStorage {
 
     private final ContentAddressedStorage target;
     private final long gcPeriodMillis;
+    // This lock is used to make new transactions block until a pending GC completes
     private final Object gcLock = new Object();
     private final ConcurrentHashMap<PublicKeyHash, AtomicInteger> openTransactions = new ConcurrentHashMap<>();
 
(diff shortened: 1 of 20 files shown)


src/peergos/server/storage/GarbageCollector.java
```diff
@@ -0,0 +1,146 @@
...
+    private int openTransactions() {
+        int res = 0;
+        for (AtomicInteger open : openTransactions.values()) {
+            res += Math.max(0, open.get());
+        }
+        return res;
+    }
+
+    public void run() {
+        while (true) {
```
cboddy: Can you add a way to signal this runnable to finish?

src/peergos/server/storage/GarbageCollector.java
```diff
@@ -0,0 +1,146 @@
...
+import java.util.concurrent.atomic.*;
+import java.util.logging.*;
+
+public class GarbageCollector implements ContentAddressedStorage {
+
+    private static final long MAX_WAIT_FOR_TRANSACTION_MILLIS = 10_000;
+
+    private final ContentAddressedStorage target;
+    private final long gcPeriodMillis;
+    private final Object gcLock = new Object();
```
cboddy: Can you add a comment about what state this lock is guarding?

ianopolous: done

src/peergos/server/storage/GarbageCollector.java
ianopolous: It doesn't need to be cleanly shutdown

src/peergos/server/storage/GarbageCollector.java
```diff
@@ -0,0 +1,148 @@
...
+                        if ((System.nanoTime() - start) / 1_000_000 > MAX_WAIT_FOR_TRANSACTION_MILLIS)
+                            openTransactions.clear();
+                        System.out.println("GC sleeping, " + openTransactions() + " open transactions");
+                        Thread.sleep(100);
+                    }
+                    long ready = System.nanoTime();
+                    target.gc().join();
+                    long done = System.nanoTime();
+                    Logging.LOG().info(String.format("GC took: %d ms waiting to start, %d ms in actual GC",
+                            (ready - start)/1000000, (done - ready)/1000000));
```
kevodwyer: add underscores to 1000000 for readability

ianopolous: done

PR #550, opened 2019-08-26
[cli] ProgressBar is now  thread-safe

cboddy:
Since the progress-updater is called concurrently the progress-bar should be thread-safe - please check it carefully. 

This also simplifies the ProgressBar.

Technically ``animationPos``is not  guarded but I don't think it matters.


ianopolous:
Why does it need another thread?

PR #552, opened 2019-08-29
Cli++

cboddy:
More commands, more completers.

### DIFF
--- a/src/peergos/server/cli/CLI.java
+++ b/src/peergos/server/cli/CLI.java
@@ -31,21 +31,22 @@ public class CLI implements Runnable {
 
     private final CLIContext cliContext;
     private final FileSystem peergosFileSystem;
-    private final ListFilesCompleter remoteFilesCompleter, localFilesCompleter;
+    private final ListFilesCompleter remoteFilesCompleter, localFilesCompleter, remoteDirsCompleter;
     private final Completer allUsernamesCompleter, followersCompleter, pendingFollowersCompleter, processFollowRequestCompleter;
     private volatile boolean isFinished;
 
     public CLI(CLIContext cliContext) {
         this.cliContext = cliContext;
         this.peergosFileSystem = new PeergosFileSystemImpl(cliContext.userContext);
-        this.remoteFilesCompleter = new ListFilesCompleter(this::remoteFilesLsFiles);
+        this.remoteFilesCompleter = new ListFilesCompleter(path -> this.remoteFilesLsFiles(path, false));
+        this.remoteDirsCompleter = new ListFilesCompleter(path -> this.remoteFilesLsFiles(path, true));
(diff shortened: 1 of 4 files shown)


src/peergos/server/cli/CLI.java
```diff
@@ -360,7 +365,37 @@ public String buildPrompt() {
...
+        try {
+            Path path = resolveToPath(pathArgument).toAbsolutePath();
+            if (path.toFile().isFile())
+                return Arrays.asList(path.toString());
+            if (path.toFile().isDirectory())
+                return Files.list(path)
+                        .map(Path::toString)
+                        .collect(Collectors.toList());
+
+            if (!path.getParent().toFile().isDirectory())
```
ianopolous: when is something both not a file and not a directory?

src/peergos/server/cli/Command.java
```diff
@@ -2,33 +2,45 @@
...
+    rm("Remove a remote-file", "rm remote-path", Arg.REMOTE),
     space("Show used remote space"),
     get_follow_requests("Show the users that have sent you a follow request"),
-    follow("Send a follow-request to another user.", "follow username-to-follow"),
-    share_read("Grant read access for a file to another user.", "share_read path <user>"),
+    follow("Send a follow-request to another user.", "follow username-to-follow", Arg.USERNAME),
+    share_read("Grant read access for a file to another user.", "share_read path <user>", Arg.REMOTE, Arg.FOLLOWER),
     passwd("Update your password"),
-    cd("change (remote) directory", "cd <path>"),
+    cd("change (remote) directory", "cd <path>", Arg.REMOTE),
```
ianopolous: would it make sense to split arg.Remote into dirOnly for this command? Otherwise an autcomplete in a dir with 1 subdir and 1000s of files will be awkward?

ianopolous:
I like where this is going! 2 minor comments.

src/peergos/server/cli/CLI.java
cboddy: Yes - if it  a partial path (ie. user hasn't finished typing it)

src/peergos/server/cli/Command.java
cboddy: Sure, I  think it's  a nice-to-have though  -  thousands of files in one dir is always going to be awkward.

cboddy:
I think this is GTG now - any other comments?

src/peergos/server/cli/Command.java
ianopolous: I think it's an important UX improvement (bash does it). It matters even is there is only 1 dir and 1 file. It's also easy to do.

cboddy: I didn't really read this the first time - yes that makes sense - now done :)

PR #557, opened 2019-09-08
[native-image] standalone binary distribution

cboddy:
Added script for building Peergos as a standalone binary with the
graalvm/native-image tool.

Also since it's quite abstruse another script for  installing that tool.


ianopolous:
Running the install script gives me,
> ./scripts/install-native-image.sh: line 17: export: `native-image=/home/ian/peergos/graalvm-ce-19.2.0/jre/lib/svm/bin/native-image': not a valid identifier

ianopolous:
First build attempt was missing a dep. Would be good to list these somewhere:
Error: Basic header file missing (<zlib.h>). Make sure headers are available on your system.

ianopolous:
Dep command (only 1 was missing):
> sudo apt install zlib1g-dev

ianopolous:
I tried running it. It get's as far as installing ipfs, but then silently terminates before ever running ipfs.

ianopolous:
It was "silent" because it ignored our logToConsol flag, but the error was in the log:
> com.oracle.svm.core.jdk.UnsupportedFeatureError: Accessing an URL protocol that was not enabled. The URL protocol http is supported but not enabled by default. It must be enabled by adding the -H:EnableURLProtocols=http option to the native-image command.

ianopolous:
Now it successfully starts ipfs! Then it fails trying to load sqlite:
java.sql.SQLException: No suitable driver found for jdbc:sqlite:/home/ian/.peergos/mutable.sql
	at java.sql.DriverManager.getConnection(DriverManager.java:689)
I suspect it is trying to use reflection to load the class, so we might need to whitelist it.

ianopolous:
I still can't get the sqlite driver to load. Even if I explicitly change the code to not use reflection with:
> DriverManager.registerDriver(new JDBC());

Which is what the sqlite static initializer does.

ianopolous:
I've also tried adding the following but it doesn't change the result
> --initialize-at-build-time=org.sqlite.JDBC

> java.sql.SQLException: No suitable driver found for jdbc:sqlite:/home/ian/.peergos/mutable.sql
	at java.sql.DriverManager.getConnection(DriverManager.java:689)
	at java.sql.DriverManager.getConnection(DriverManager.java:270)
	at peergos.server.corenode.JdbcIpnsAndSocial.buildSqlLite(JdbcIpnsAndSocial.java:243)

PR #579, opened 2019-10-05
Check validity of target WrierData on pointer writes

ianopolous:


src/peergos/server/util/PeergosNetworkUtils.java
```diff
@@ -26,7 +26,7 @@
 public class PeergosNetworkUtils {
 
     public static String generateUsername(Random random) {
-        return "username_" + Math.abs(random.nextInt() % 1_000_000);
+        return "username_" + Math.abs(random.nextInt() % 1_000_000_000);
```
cesarosum: Why this number?

ianopolous: Because it's bigger, round, and less than max int.

PR #584, opened 2019-10-27
Total rewrite of revoking write access

ianopolous:
This is a fairly large and fundamental change and worth understanding in detail.

We weren't handling nested write access when revoking write access. This fixes that. 

The rewrite makes revoking write access a single pass over the subtree rather than the previous 3. 
It also makes it atomic as far as the filesystem view is concerned. The operation can crash at any point and the filesystem will be in a consistent state, without any data loss. (There may be duplicated data lying around, but this can be handled in a future pr with a similar mechanism to file uploads - by first writing a transaction file describing the operation.)

We have to be careful when encountering descendants with different signing keys to first auth the new signing key by adding it as an owned key to the new parent signer, and similarly clean up old signers as we go.  

I've also added a path based lookup to who something has been shared with (as well as the mapkey based one). This allows easy resharing of all subtree shares to remaining sharees when revoking access to others. 

Granting write access is now also O(1) if the parent already has a different signer (previously we always rotated the child).

The critical method where everything happens is CryptreeNode.rotateAllKeys()

One thing to bear in mind is that subsequent chunks of files or dirs do not have a parent link (otherwise we'd have to keep them updated)

UPDATE:
After submitting this I found a pre-existing issue with granting write access (and made a test for it). I've fixed that and made it atomic by making it use the same method revoking access uses to rotate all the keys of a subtree. 

UPDATE2:
The other interesting change was fixing TrieNodeImpl to handle revoked capabilities correctly.

### DIFF
--- a/src/peergos/gwt/emu/java/nio/file/Path.java
+++ b/src/peergos/gwt/emu/java/nio/file/Path.java
@@ -43,6 +43,18 @@ public class Path {
         return new Path(pathString + "/" + other);
     }
 
+    public Path resolve(Path other) {
+        return resolve(other.pathString);
+    }
+
+    public boolean isAbsolute() {
+        return pathString.startsWith("/");
+    }
+
+    public boolean startsWith(Path other) {
+        return pathString.startsWith(other.pathString);
+    }
+
     public File toFile() {
         throw new IllegalArgumentException("Not implemented!");
     }
--- a/src/peergos/gwt/emu/java/nio/file/Path.java
+++ b/src/peergos/gwt/emu/java/nio/file/Path.java
@@ -1,6 +1,7 @@
 package java.nio.file;
 
 import java.io.File;
+import java.util.*;
 
 public class Path {
 
@@ -71,4 +72,17 @@ public class Path {
         throw new IllegalArgumentException("Not implemented!");
     }
 
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Path path = (Path) o;
+        return Objects.equals(pathString, path.pathString);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(pathString);
+    }
 }
--- a/src/peergos/server/tests/MultiUserTests.java
+++ b/src/peergos/server/tests/MultiUserTests.java
@@ -106,6 +106,11 @@ public class MultiUserTests {
     }
 
     @Test
+    public void shareAFileWithDifferentSigner() {
+        PeergosNetworkUtils.shareFileWithDifferentSigner(network, network, 1, random);
+    }
+
+    @Test
     public void grantAndRevokeDirReadAccess() throws Exception {
         PeergosNetworkUtils.grantAndRevokeDirReadAccess(network, network, 2, random);
     }
--- a/src/peergos/server/tests/MultiUserTests.java
+++ b/src/peergos/server/tests/MultiUserTests.java
@@ -121,6 +121,11 @@ public class MultiUserTests {
     }
 
     @Test
+    public void grantAndRevokeDirWriteAccessWithNestedWriteAccess() {
+        PeergosNetworkUtils.grantAndRevokeDirWriteAccessWithNestedWriteAccess(network, random);
+    }
+
+    @Test
     public void safeCopyOfFriendsReadAccess() throws Exception {
         TriFunction<UserContext, UserContext, String, CompletableFuture<Boolean>> readAccessSharingFunction =
                 (u1, u2, filename) ->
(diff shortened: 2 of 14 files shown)


ianopolous:
I think this is good to go now.

cesarosum:
Was the problem that `gwtc` was giving an OOM error?

ianopolous:
No there just some new methods on Path that we were using that weren't in the gwt version of Path

cesarosum:
Ah I was getting an OOM error. I `git pull`'d and it's compiling fine now.

cesarosum:
I don't think this is working. I've compiled and am running a Peergos server on localhost with three users and can't get sharing to work - even read only is failing.

ianopolous:
Yeah, that won't work yet, there's an extra parameter (the path) which the JS needs to send in the sharing calls.

cesarosum:
Ah OK. Is there a new branch in `web-ui` for testing with this?

ianopolous:
not yet ;-)

cesarosum:
OK looks like I'll have to actually read some code then :-p

ianopolous:
fix/update-to-new-sharing-api is the branch in web ui - just pushed

src/peergos/gwt/emu/java/nio/file/Path.java
```diff
@@ -43,6 +43,18 @@ public Path resolve(String other) {
         return new Path(pathString + "/" + other);
     }
 
+    public Path resolve(Path other) {
```
kevodwyer: Could possibly check for null here and in a few other places in this class.

src/peergos/server/tests/MultiUserTests.java
```diff
@@ -120,6 +120,16 @@ public void grantAndRevokeNestedDirWriteAccess() {
         PeergosNetworkUtils.grantAndRevokeNestedDirWriteAccess(network, random);
     }
 
+    @Test
+    public void grantAndRevokeNestedDirWriteAccess2() {
```
kevodwyer: Better method name

src/peergos/shared/user/SharedWithCache.java
```diff
@@ -15,11 +14,40 @@
     //K,V
     // K = mapKey of absolute capability
     // V = list of usernames the file/dir is shared with
-    private Map<ByteArrayWrapper, Set<String>> sharedWithReadAccessCache = new ConcurrentHashMap<>();
-    private Map<ByteArrayWrapper, Set<String>> sharedWithWriteAccessCache = new ConcurrentHashMap<>();
+    private final Map<ByteArrayWrapper, Set<String>> sharedWithReadAccessCache = new ConcurrentHashMap<>();
+    private final Map<ByteArrayWrapper, Set<String>> sharedWithWriteAccessCache = new ConcurrentHashMap<>();
+    // These would be more space efficient in a trie
+    private final Map<Path, Set<String>> writeShares = new ConcurrentHashMap<>();
```
kevodwyer: What about files that are moved?

src/peergos/server/tests/MultiUserTests.java
ianopolous: I agree, but couldn't think of one. Any ideas?

ianopolous: done

src/peergos/gwt/emu/java/nio/file/Path.java
ianopolous: a null pointer is fine I think

src/peergos/shared/user/SharedWithCache.java
kevodwyer: I just tested and your are right the current solution does not work with move. I'm sure it used to - hence the key of the map was not a path.

ianopolous: What is the desired behaviour?

kevodwyer: If i move directories around, i would find it surprising that my shares stopped working. This is especially the case if i move a parent folder to a shared folder/files. Maybe revost at a later date.

ianopolous: Agreed, we should think more about the situations here and what is most intuitive. For a different PR.

ianopolous:
The last commit fixing granting write access has broken a few tests so I'm trying to fix those...

ianopolous:
I'm hoping all tests are green now. So I'm going to do some more manual testing then hopefully get one more review before merging today.

ianopolous:
All the tests are green but there appear to be issues in the webui. Revoke read or write seems to not work at all, unless in debugger for example.

ianopolous:
This is definitely good to go now. All is good in the webui

PR #592, opened 2019-11-15
Implement postgres as an option for a database

ianopolous:


src/peergos/server/corenode/JdbcIpnsAndSocial.java
```diff
@@ -112,35 +142,38 @@ public boolean delete() {
...
-            String tableName = rs.getString("name");
-            missingTables.remove(tableName);
+            String tableName = rs.getString(1);
+            tables.add(tableName);
         }
 
-        for (String missingTable: missingTables) {
-            try {
+        try {
+            if (! tables.contains("followrequests")) {
```
cboddy: A few suggestions:
* arguably the DDL should live in a separate script (for both implementations), particularly since there is more to do now with the postgres impl. around user  setup that cannot be done in this client. 
*  if we do keep these DDL statement here we  can remove this conditional since both sqlite and  postgres support ``create table if not exists`` statements.

src/peergos/server/util/Postgres.java
```diff
@@ -0,0 +1,21 @@
...
+    public static Connection build(String host,
+                                   int port,
+                                   String database,
+                                   String user,
+                                   String password) throws SQLException {
+        Driver driver = new Driver();
+        Properties props = new Properties();
+        props.setProperty("user", user);
+        props.setProperty("password", password);
+        return driver.connect("jdbc:postgresql://" + host + ":" + port + "/" + database, props);
```
cboddy: Just to flag: the traffic on this connection will be in the clear (and if  we want to use make use of the native Postgres SSL connection then  we need  `&ssl=true`  in the query string.

src/peergos/server/corenode/JdbcIpnsAndSocial.java
ianopolous: create table if not exists would be great! I didn't know that existed

src/peergos/server/util/Postgres.java
ianopolous: Yep anyone using this should know to use a VPN, but I'll add a warning.

PR #595, opened 2019-11-17
Cleaner idempotent table creation

ianopolous:


cboddy:
which version are you  running against? It has been a feature since `9.1`

https://www.postgresql.org/docs/current/sql-createtable.html

ianopolous:
The latest. It's all good now. ;-)

PR #599, opened 2019-12-01
[aggregated-metrics]

cboddy:
This change adds globally-aggregated metrics for some of the
Peergos-server calls.

It is implemented with Prometheus using the HTTP exporter, and adds
the minimum dependecy jars needed.

This is a jumping off point - we can add/remove more of these stats as needed. Also note the exporter is not enabled by default.


src/peergos/server/AggregatedMetrics.java
```diff
@@ -0,0 +1,40 @@
...
+    public static final Counter MUTABLE_POINTERS_SET  = build("mutable_pointers_set", "Total mutable-pointers set calls.");
+    public static final Counter MUTABLE_POINTERS_GET  = build("mutable_pointers_get", "Total mutable-pointers get calls.");
+
+    public static final Counter GET_PUBLIC_KEY  = build("core_node_get_public_key", "Total get-public-key calls.");
+    public static final Counter GET_PUBLIC_KEY_CHAIN  = build("core_node_get_chain", "Total get-public-key-chain calls.");
+    public static final Counter UPDATE_PUBLIC_KEY_CHAIN  = build("core_node_update_chain", "Total getupdate-public-key-chain calls.");
+
+
+    public static void startExporter(int port) throws IOException {
+        HTTPServer server = new HTTPServer(port);
```
ianopolous: Just need to clarify exactly what this is listening on.

It ends up here: https://docs.oracle.com/javase/7/docs/api/java/net/InetSocketAddress.html#InetSocketAddress(int)
It says: The wildcard is a special local IP address. It usually means "any"

cboddy: Yep, good spot  - it's now configurable with default "localhost".

PR #601, opened 2019-12-04
add java 11 support via gwt 2.8.3 custom build

kevodwyer:
see BUILD-GWT.txt in gwt folder for instructions on how to build


cboddy:
> Files changed 6,939
o_0

ianopolous:
I've tested a bunch of stuff. LGTM! The only problem now is the directories in the jar have acquired a timestamp, which breaks reproducible builds.

ianopolous:
That problem with dir timestamps seems to have happened earlier. So this is good to go.

ianopolous:
Just need to figure out why test classpath is broken in java 11 before merging this.

ianopolous:
hmm java thinks the output jar from this is invalid for some reason, but it opens fine as a zip

PR #604, opened 2019-12-10
Fix cpu maxing by logging threads for ipfs sub process

ianopolous:
(Only happens on java 11 bizarrely)


ianopolous:
I think the conclusion is this doesn't fix anything. The problem is deeper in the JVM/native libraries.

ianopolous:
The problem was not in our code, but deeper in the JVM. Switching to debian's build of openjdk fixed it.

Issue #610, opened 2019-12-31
bug: revoking read-access on a  folder in which a subfolder is write-shared


cboddy:
(Found in the fuzzer)

steps to reproduce (pseudo-code):

create  two users: left, right
mkdir /right/2/4
write-share /right/2/4 
read-share /right/2
revoke-read-share /right/2 

*BOOM*

PR #612, opened 2019-12-31
Fix issue found by fuzzer.

ianopolous:
Revoking read access to a dir, when write access had been granted to a
child was failing.


ianopolous:
Fixes https://github.com/Peergos/Peergos/issues/610

PR #615, opened 2020-01-02
Implement and integrate a direct S3 block store

ianopolous:
It still needs to be tested, but should work as is, and be compatible with ipfs.

GC has to be disabled. Eventually we can write a simple GC implementation that works in conjunction with the sql based transaction store.


src/peergos/server/storage/S3BlockStorage.java
```diff
@@ -0,0 +1,335 @@
...
+
+    private static final Logger LOG = Logger.getGlobal();
+
+    private static final Histogram readTimerLog = Histogram.build()
+            .labelNames("filesize")
+            .name("block_read_seconds").help("Time to read a block from immutable storage")
+            .exponentialBuckets(0.01, 2, 16).register();
+    private static final Histogram writeTimerLog = Histogram.build()
+            .labelNames("filesize")
+            .name("s3_block_write_seconds").help("Time to read a block from immutable storage")
```
kevodwyer: help message says read rather than write

src/peergos/server/storage/S3BlockStorage.java
```diff
@@ -0,0 +1,335 @@
...
+
+    @Override
+    public CompletableFuture<Multihash> id() {
+        return Futures.of(id);
+    }
+
+    @Override
+    public CompletableFuture<TransactionId> startTransaction(PublicKeyHash owner) {
+        // TODO delegate this to a sql table
+        return CompletableFuture.completedFuture(new TransactionId(Long.toString(System.currentTimeMillis())));
```
kevodwyer: Risk of duplicate transactionId. Resolution of currentTimeMillis is very low.

ianopolous: That's okay until we have a sql version because they don't do anything.

src/peergos/server/storage/S3BlockStorage.java
ianopolous: Fixed

src/peergos/server/storage/JdbcTransactionStore.java
```diff
@@ -0,0 +1,103 @@
...
+
+import java.sql.*;
+import java.util.*;
+import java.util.logging.*;
+import java.util.stream.*;
+
+public class JdbcTransactionStore implements TransactionStore {
+	private static final Logger LOG = Logging.LOG();
+
+    private static final String INSERT_TRANSACTIONS_BLOCK = "INSERT INTO transactions (tid, owner, hash) VALUES(?, ?, ?);";
```
cboddy: `pending_transaction` would be a better name for this table imho.

src/peergos/server/sql/SqliteCommands.java
```diff
@@ -0,0 +1,15 @@
+package peergos.server.sql;
+
+public class SqliteCommands implements SqlSupplier {
```
cboddy: As discussed I think we should  look to  remove any DDL  from these interfaces/implementations, which amounts to removing the `SqlSupplier` interface I think. We can do this in a  separate PR.

src/peergos/server/storage/TransactionStore.java
```diff
@@ -0,0 +1,19 @@
+package peergos.server.storage;
+
+import peergos.shared.crypto.hash.*;
+import peergos.shared.io.ipfs.multihash.*;
+import peergos.shared.storage.*;
+
+import java.util.*;
+
+public interface TransactionStore {
```
cboddy: Really nice.

src/peergos/server/storage/JdbcTransactionStore.java
```diff
@@ -0,0 +1,103 @@
+package peergos.server.storage;
```
cboddy: Could this get some explicit test coverage (using sqlite)?

src/peergos/server/Main.java
```diff
@@ -104,6 +104,7 @@
                     new Command.Arg("mutable-pointers-file", "The filename for the mutable pointers datastore", true, "mutable.sql"),
                     new Command.Arg("social-sql-file", "The filename for the follow requests datastore", true, "social.sql"),
                     new Command.Arg("space-requests-sql-file", "The filename for the space requests datastore", true, "space-requests.sql"),
+                    new Command.Arg("transactions-sql-file", "The filename for the transactions datastore", false, "transactions.sql"),
```
cboddy: As before, I would advocate merging the (three?) sqlite files.

src/peergos/server/storage/S3BlockStorage.java
```diff
@@ -0,0 +1,337 @@
...
+            .name("block_read_seconds").help("Time to read a block from immutable storage")
+            .exponentialBuckets(0.01, 2, 16).register();
+    private static final Histogram writeTimerLog = Histogram.build()
+            .labelNames("filesize")
+            .name("s3_block_write_seconds").help("Time to read a block from immutable storage")
+            .exponentialBuckets(0.01, 2, 16).register();
+
+    private final Multihash id;
+    private final AmazonS3 s3Client;
+    private final String bucket, folder;
```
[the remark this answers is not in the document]
ianopolous: I've confirmed the library is thread safe. We will only have one instance. Folder is just the folder in the bucket where everything is stored.

src/peergos/server/storage/S3Config.java
```diff
@@ -0,0 +1,29 @@
...
+    }
+    public static boolean useS3(Args a) {
+        return a.hasArg("s3.bucket");
+    }
+
+    public static S3Config build(Args a) {
+        String path = a.getArg("s3.path", "blocks");
+        String bucket = a.getArg("s3.bucket");
+        String region = a.getArg("s3.region");
+        String accessKey = a.getArg("s3.accessKey", "");
```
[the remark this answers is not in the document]
ianopolous: Then the aws library tries to get the config from ~/.aws/config

src/peergos/server/Main.java
```diff
@@ -376,29 +393,18 @@ public static UserService startPeergos(Args a) {
                 boolean enableGC = a.getBoolean("enable-gc", false);
                 if (enableGC)
                     throw new IllegalStateException("GC has not been implemented when directly using S3!");
-                localDht = new S3BlockStorage(s3Config, id);
+                Connection transactionsDb = usePostgres ?
+                    database :
+                    Sqlite.build(Sqlite.getDbPath(a, "transactions-sql-file"));
```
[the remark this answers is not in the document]
ianopolous: Each one is locked independently. It also makes it trivial to delete one for example, which I occasionally do.

src/peergos/server/storage/JdbcTransactionStore.java
ianopolous: There isn't a non pending transaction.

src/peergos/server/storage/S3BlockStorage.java
```diff
@@ -0,0 +1,337 @@
...
+            return success.apply(hash);
+        } catch (AmazonServiceException e) {
+            /* Caught an AmazonServiceException,
+               which means our request made it
+               to Amazon S3, but was rejected with an error response
+               for some reason.
+            */
+            if ("NoSuchKey".equals(e.getErrorCode())) {
+                Histogram.Timer readTimer = readTimerLog.labels("0").startTimer();
+                readTimer.observeDuration();
```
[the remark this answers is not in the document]
ianopolous: This counts reads for non existent blocks

src/peergos/server/storage/S3BlockStorage.java
```diff
@@ -0,0 +1,337 @@
...
+        if (hash instanceof Cid && ((Cid) hash).codec == Cid.Codec.Raw)
+            throw new IllegalStateException("Need to call getRaw if cid is not cbor!");
+        return getRaw(hash).thenApply(opt -> opt.map(CborObject::fromByteArray));
+    }
+
+    @Override
+    public CompletableFuture<Optional<byte[]>> getRaw(Multihash hash) {
+        return Futures.of(map(hash, h -> {
+            GetObjectRequest get = new GetObjectRequest(bucket, folder + hashToKey(hash));
+            Histogram.Timer readTimer = readTimerLog.labels(h.toString()).startTimer();
```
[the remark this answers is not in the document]
ianopolous: Great catch!

src/peergos/server/storage/JdbcTransactionStore.java
ianopolous: Done

PR #618, opened 2020-01-10
Extract out UsageStore interface from SpaceCheckingKeyFilter

ianopolous:
It will probably change again once a sql version is implemented, e.g. to use CASes rather than writes.


src/peergos/server/Main.java
```diff
@@ -433,8 +433,10 @@ public static UserService startPeergos(Args a) {
             JdbcSpaceRequests spaceRequests = JdbcSpaceRequests.build(spaceDb, sqlCommands);
             UserQuotas userQuotas = new UserQuotas(quotaFilePath, defaultQuota, maxUsers);
             CoreNode signupFilter = new SignUpFilter(core, userQuotas, nodeId);
+            RamUsageStore usageStore = RamUsageStore.build(statePath, localPointers, localDht);
+            SpaceCheckingKeyFilter.update(usageStore, localPointers, localDht);
```
kevodwyer: This update method 'feels' wrong.  How about turning into a builder method?

src/peergos/server/space/SpaceCheckingKeyFilter.java
```diff
@@ -0,0 +1,319 @@
...
+        usageStore.addWriter(event.username, event.keyHash);
+        ForkJoinPool.commonPool().submit(() -> processCorenodeEvent(event.username, event.keyHash));
+    }
+
+    /** Update our view of the world because a user has changed their public key (or registered)
+     *
+     * @param username
+     * @param writer
+     */
+    public void processCorenodeEvent(String username, PublicKeyHash writer) {
```
kevodwyer: set visibility to private?

src/peergos/server/Main.java
ianopolous: I get what you're saying, but it is updating the calculated usage in case any of the pointers have changed.

src/peergos/server/space/SpaceCheckingKeyFilter.java
ianopolous: done

ianopolous:
The UsageStore interface ended up being more complicated than I'd hoped. any ideas for simplifying it before I dive into a sql version?

kevodwyer:
> The UsageStore interface ended up being more complicated than I'd hoped. any ideas for simplifying it before I dive into a sql version?

small thing: For the interface, I would re-order the methods and group them.

PR #619, opened 2020-01-12
this became a huge bottleneck when testing async upload

kevodwyer:


src/peergos/shared/cbor/CborEncoder.java
```diff
@@ -386,15 +386,13 @@ protected void writeSimpleType(int majorType, int value) throws IOException {
...
+     * @param bytes the byte string to write, can be <code>null</code> in which case a byte-string of length <tt>0</tt> is written.
      * @throws IOException in case of I/O problems writing the CBOR-encoded value to the underlying output stream.
      */
     protected void writeString(int majorType, byte[] bytes) throws IOException {
         int len = (bytes == null) ? 0 : bytes.length;
         writeType(majorType, len);
-        for (int i = 0; i < len; i++) {
-            m_os.write(bytes[i]);
-        }
+        m_os.write(bytes);
```
ianopolous: what about the null case?

PR #622, opened 2020-01-14
improve upload file method called from JS

kevodwyer:
used in file-upload-improvements web-ui branch.


ianopolous:
What's the context for this? What triggers it?

ianopolous:
There seem to be a few test failures.

PR #623, opened 2020-01-14
[simulate-sharing] WIP: add read and write shared-files to as ops to …

cboddy:
…the simulator


ianopolous:
Doesn't seem to be compiling.

PR #626, opened 2020-01-27
Cryptree links

ianopolous:
This implements an improved mechanism for granting write access. 

This is done by a new type of cryptree node - a link node. A link node looks like a directory node with exactly one child, and with a isLink flag set on its metadata. This allows us to separate granting write access from granting rename permission. The implementation changes granting write access to add a link node in the same signing space as the parent directory, which points to the file in the new signing space. When retrieving the file we always use the name from the link node's metadata, and the remaining properties from the actual target node's metadata. 

We could use this same mechanism to restrict other permissions that are controlled by the metadata, for example changing the file size. Or even make a file append only by inserting the link node part way through the file. We also use them to implement the normal concept of a soft/hard links. 

After this PR we regain the usual filesystem semantics that if you only have write access to a file/dir and not its parent then you can't rename it. 

Because the only difference between a link node and a different type of cryptree node is a flag in the encrypted metadata, the network cannot distinguish between a link node, a directory, or a small file.

### DIFF
--- a/src/peergos/shared/user/fs/FileWrapper.java
+++ b/src/peergos/shared/user/fs/FileWrapper.java
@@ -139,6 +139,10 @@ public class FileWrapper {
         return pointer;
     }
 
+    public RetrievedCapability getLinkPointer() {
+        return linkPointer.get();
+    }
+
     public boolean isRoot() {
         return props.name.equals("/");
     }
@@ -376,6 +380,10 @@ public class FileWrapper {
     public boolean isDirectory() {
         boolean isNull = pointer == null;
         return isNull || pointer.fileAccess.isDirectory();
+    }
+
+    public boolean isLink() {
+        return props.isLink;
     }
 
     @JsMethod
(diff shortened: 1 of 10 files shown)


src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -948,12 +976,16 @@ static boolean isLegalName(String name) {
                                                  FileWrapper parent,
                                                  boolean overwrite,
                                                  UserContext userContext) {
-        setModified();
         if (! isLegalName(newFilename))
             return CompletableFuture.completedFuture(parent);
+        if (! parent.isWritable())
+            return Futures.errored(new IllegalStateException("You cannot rename something without write access to the parent!"));
```
kevodwyer: Drop the 'you' from start of sentence. Use 'Unable to rename..'

ianopolous: Done

PR #644, opened 2020-03-18
video capture

kevodwyer:


src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -717,8 +717,11 @@ public static int getNumberOfChunks(long size) {
     }
 
     private CompletableFuture<Snapshot> recalculateThumbnail(Snapshot snapshot, Committer committer, String filename, AsyncReader fileData
-             , boolean isHidden, long fileSize, NetworkAccess network, WritableAbsoluteCapability fileWriteCap, Optional<byte[]> streamSecret
+             , boolean isHidden, long startIndex, long fileSize, NetworkAccess network, WritableAbsoluteCapability fileWriteCap, Optional<byte[]> streamSecret
     ) {
+        if(startIndex != 0) {
```
ianopolous: So in this case you're not recalculating the thumbnail? If so, then why not just avoid calling this method at all?

kevodwyer: have refactored

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -717,7 +725,7 @@ public static int getNumberOfChunks(long size) {
     }
 
     private CompletableFuture<Snapshot> recalculateThumbnail(Snapshot snapshot, Committer committer, String filename, AsyncReader fileData
-             , boolean isHidden, long fileSize, NetworkAccess network, WritableAbsoluteCapability fileWriteCap, Optional<byte[]> streamSecret
+             , boolean isHidden, long startIndex, long fileSize, NetworkAccess network, WritableAbsoluteCapability fileWriteCap, Optional<byte[]> streamSecret
```
ianopolous: no need for new parameter now

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -654,24 +654,32 @@ public static int getNumberOfChunks(long size) {
                                                     throw new IllegalStateException("File already exists with name " + filename);
                                                 FileWrapper child = childOpt.get();
                                                 FileProperties childProps = child.getFileProperties();
+
+                                                TriFunction<FileWrapper, Snapshot, AsyncReader, CompletableFuture<Snapshot>> recalculateThumbnailIfNecessary =
+                                                        (updatedChild, latestSnapshot, is) -> {
+                                                    if(startIndex != 0) {
```
ianopolous: you could use an inline statement in both places. arguable whether or not that is simpler/shorter
e.g. startIndex != 0 ? Futures.of(latestSnapshot) : updatedChild.recalculateThumbnail(...)

PR #654, opened 2020-04-08
handle exception in Futures.combine* 

kevodwyer:


kevodwyer:
In java it is fine (the extra code is not executed).
In JS the hack is required (don't know why).

ianopolous:
Ok then it sounds like it could be a bug in our CompletableFuture implementation for gwt

PR #656, opened 2020-04-09
Serializability was Causing unnecessary issues in Postgres, so revert…

ianopolous:
… to default;


src/peergos/server/space/JdbcUsageStore.java
```diff
@@ -122,29 +125,15 @@ public void confirmUsage(String username, PublicKeyHash writer, long usageDelta,
 
     @Override
     public void addPendingUsage(String username, PublicKeyHash writer, int size) {
-        int userId = getUserId(username);
         int writerId = getWriterId(writer);
-        try (Connection conn = getNonCommittingConnection();
-             PreparedStatement defaultInsert = conn.prepareStatement(commands.insertOrIgnoreCommand(
```
kevodwyer, line 128: good find on dup query

PR #658, opened 2020-04-15
Upgrade to ipfs v0.5.0

ianopolous:
This upgrades to ipfs v0.5.0-rc2

The main thing that needs reviewing is the switch to requiring http POSTs on our api, unless we are explicitly a public server. As well as making sure the JS client does use GETs when talking to a public server - in this case public means not localhost. 

The issue is that browsers allow any random site to make GETs to localhost, but not POSTs. It's not a huge deal for us compared to ipfs, because all our calls are authed anyway if they aren't nullipotent.


src/peergos/server/net/DHTHandler.java
```diff
@@ -29,28 +29,37 @@
...
     private final BiFunction<PublicKeyHash, Integer, Boolean> keyFilter;
     private final String apiPrefix;
+    private final boolean isPublicServer;
 
     public DHTHandler(ContentAddressedStorage dht,
                       Hasher hasher,
                       BiFunction<PublicKeyHash, Integer, Boolean> keyFilter,
-                      String apiPrefix) {
+                      String apiPrefix,
+                      boolean isPublicServer) {
```
kevodwyer, line 38: Other option is to have a Handler superclass. call super(isPublicServer) in constructor. Then have the super.handle() deal with checking and potentially throwing 405 before calling a new method called handleRequest that each subclass overrides.

src/peergos/server/storage/IPFS.java
```diff
@@ -505,12 +505,17 @@ private Object retrieveAndParse(String path) throws IOException {
 
     private static byte[] get(URL target) throws IOException {
         HttpURLConnection conn = (HttpURLConnection) target.openConnection();
-        conn.setRequestMethod("GET");
+        conn.setDoOutput(true);
+        conn.setRequestMethod("POST");
```
kevodwyer, line 509: probably best to add a comment and link to ipfs commit on why POST is used

ianopolous: Done

src/peergos/server/net/DHTHandler.java
ianopolous: All of these are going to be rewritten once we move to netty. I think this is simple enough for now. Then I imagine we'd have a chain of handlers, and if the first doesn't handle it, then the second does etc. i.e. an actual web framework.

PR #673, opened 2020-05-04
Fix issue with FileWrapper.getUpdated() due to missing equals() on CommittedWriterData

kevodwyer:


src/peergos/shared/user/CommittedWriterData.java
```diff
@@ -16,4 +18,18 @@ public CommittedWriterData(MaybeMultihash hash, WriterData props) {
...
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        CommittedWriterData that = (CommittedWriterData) o;
+        return Objects.equals(hash, that.hash);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(hash);
```
ianopolous, line 32: can just be hash.hashcode()

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -109,6 +109,13 @@ public FileWrapper withVersion(Snapshot version) {
         return new FileWrapper(capTrie, pointer, linkPointer, entryWriter, ownername, version);
     }
 
+    @JsMethod
+    public CompletableFuture<FileWrapper> getLatest(NetworkAccess network) {
+        return network.synchronizer.getValue(owner(), writer())
+                .thenCompose(finished -> getUpdated(finished, network));
```
ianopolous, line 115: maybe current or latest is a better name than finished?

src/peergos/shared/user/CommittedWriterData.java
ianopolous: I really miss non nullable types in Java

src/peergos/shared/user/fs/FileWrapper.java
kevodwyer: done

src/peergos/shared/user/CommittedWriterData.java
kevodwyer: fixed

PR #678, opened 2020-05-12
Implement CommittableStorage which removes the linear sequential chain

ianopolous:
of block puts after each chunk to update the champ.

It does this by evaluating the writes locally, and then pushing all the
blocks just before committing them with a mutable pointer write.

### DIFF
--- a/src/peergos/shared/storage/CommittableStorage.java
+++ b/src/peergos/shared/storage/CommittableStorage.java
@@ -20,9 +20,12 @@ public class CommittableStorage extends DelegatingStorage {
         this.target = target;
     }
 
-    public synchronized CompletableFuture<Boolean> flush() {
-        List<PutArgs> local = new ArrayList<>(pending);
-        pending.clear();
+    public CompletableFuture<Boolean> flush() {
+        List<PutArgs> local;
+        synchronized (pending) {
+            local = new ArrayList<>(pending);
+            pending.clear();
+        }
         List<CompletableFuture<Multihash>> uploads = local.stream()
                 .map(p -> target.put(p.owner, p.writer, p.signature, p.block, p.tid)
                         .thenApply(h -> {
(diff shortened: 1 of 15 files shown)


src/peergos/shared/storage/CommittableStorage.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public CompletableFuture<Multihash> put(PublicKeyHash owner,
+                                            PublicKeyHash writer,
+                                            byte[] signature,
+                                            byte[] block,
+                                            TransactionId tid) {
+        byte[] sha256 = Arrays.copyOfRange(signature, signature.length - 32, signature.length);
+        Cid cid = hashToCid(sha256, false);
+        synchronized (pending) {
```
kevodwyer, line 46: could use a synchonized List implementation via Collections.synchronizedList.

src/peergos/shared/storage/CommittableStorage.java
```diff
@@ -0,0 +1,71 @@
...
+        return new Cid(CID_V1, isRaw ? Cid.Codec.Raw : Cid.Codec.DagCbor, Multihash.Type.sha2_256, sha256);
+    }
+
+    private static class PutArgs {
+        public final PublicKeyHash owner, writer;
+        public final Multihash expected;
+        public final byte[] signature, block;
+        public final TransactionId tid;
+
+        public PutArgs(PublicKeyHash owner, PublicKeyHash writer, Multihash expected, byte[] signature, byte[] block, TransactionId tid) {
```
kevodwyer, line 62: not sure  'expected' var name is descriptive

ianopolous: Expected because it is what we have calculated, it is what we expect the server to return.

src/peergos/shared/storage/CommittableStorage.java
ianopolous: I've reduced the synchronized block. Now it's pretty close to optimal I think.

PR #680, opened 2020-05-13
add Retry content address storage implementation

kevodwyer:
There is a corresponding web-ui PR


src/peergos/shared/storage/RetryStorage.java
```diff
@@ -0,0 +1,160 @@
...
+
+
+    public <Y> CompletableFuture<Y> runWithRetry(Supplier<CompletableFuture<Y>> func) {
+        CompletableFuture<Y> res = new CompletableFuture<>();
+        Function<Integer, CompletableFuture<Y>> compose = recurse(
+            (i, f) -> {
+                func.get()
+                    .thenAccept(res::complete)
+                    .exceptionally(e ->  {
+                        if(i == 3) {
```
ianopolous, line 53: Can we make the number of retries configurable per instance of this class?

src/peergos/shared/storage/RetryStorage.java
```diff
@@ -0,0 +1,160 @@
...
+
+    private static <T, U> Function<T, U> recurse(Recursable<T, U> f) {
+        return t -> f.apply(t, f);
+    }
+
+
+    public <Y> CompletableFuture<Y> runWithRetry(Supplier<CompletableFuture<Y>> func) {
+        CompletableFuture<Y> res = new CompletableFuture<>();
+        Function<Integer, CompletableFuture<Y>> compose = recurse(
+            (i, f) -> {
```
ianopolous, line 49: Why do we need f as a parameter?

PR #693, opened 2020-06-13
merge sharing screens

kevodwyer:
web-ui PR: https://github.com/Peergos/web-ui/pull/156
- getLatest() does not work when files are shared
- share*AccessWith and sharedWith now take a string path. This ensures the latest version of file is retrieved.

### DIFF
--- a/src/peergos/gwt/emu/java/nio/file/Path.java
+++ b/src/peergos/gwt/emu/java/nio/file/Path.java
@@ -40,7 +40,12 @@ public class Path {
             return new Path(pathString);
         } else {
             if(index == pathString.length() - 1) {
-                return null;
+                if (index == 0) {
+                    return null;
+                } else {
+                    int secondLastIndex = pathString.lastIndexOf(SEPARATOR, index - 1);
+                    return new Path(pathString.substring(secondLastIndex + 1, index));
+                }
             } else {
                 return new Path(pathString.substring(index + 1));
             }
--- a/src/peergos/gwt/emu/java/nio/file/Path.java
+++ b/src/peergos/gwt/emu/java/nio/file/Path.java
@@ -35,7 +35,16 @@ public class Path {
     }
 
     public Path getFileName() {
-        throw new IllegalArgumentException("Not implemented!");
+        int index = pathString.lastIndexOf(SEPARATOR);
+        if (index == -1) {
+            return new Path(pathString);
+        } else {
+            if(index == pathString.length() - 1) {
+                return null;
+            } else {
+                return new Path(pathString.substring(index + 1));
+            }
+        }
     }
 
     public Path resolve(String other) {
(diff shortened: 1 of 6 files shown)


src/peergos/gwt/emu/java/nio/file/Path.java
```diff
@@ -35,7 +35,16 @@ public Path getParent() {
     }
 
     public Path getFileName() {
-        throw new IllegalArgumentException("Not implemented!");
+        int index = pathString.lastIndexOf(SEPARATOR);
+        if (index == -1) {
+            return new Path(pathString);
+        } else {
+            if(index == pathString.length() - 1) {
+                return null;
```
ianopolous, line 43: why null?

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -109,13 +109,6 @@ public FileWrapper withVersion(Snapshot version) {
         return new FileWrapper(capTrie, pointer, linkPointer, entryWriter, ownername, version);
     }
 
-    @JsMethod
-    public CompletableFuture<FileWrapper> getLatest(NetworkAccess network) {
```
ianopolous, line 113: aren't we going to need this in other features?

src/peergos/gwt/emu/java/nio/file/Path.java
kevodwyer: That is what the Path method does also.

src/peergos/shared/user/fs/FileWrapper.java
kevodwyer: For video capture - yes. But as it doesn't seem to work when the file is subsequently shared it seemed prudent to remove

src/peergos/gwt/emu/java/nio/file/Path.java
ianopolous: It only returns null if the path it empty: https://docs.oracle.com/javase/7/docs/api/java/nio/file/Path.html#getFileName()

We are different on e.g. /somedir/

src/peergos/shared/user/UserContext.java
```diff
@@ -1374,15 +1374,19 @@ private static void ensureAllowedToShare(FileWrapper file, String ourname, boole
         });
     }
 
-    public CompletableFuture<Boolean> shareWriteAccessWithAll(Path fileToShare,
-                                                               Set<String> writersToAdd) {
+
+    @JsMethod
+    public CompletableFuture<Boolean> shareWriteAccessWith(String pathToFile,
```
ianopolous, line 1379: In general a typed api is better than a string api. So I'd prefer if the JS method called the typed api. I'm hoping at somepoint we can remove all these JS only apis into a sub class or something. Java code for sure should be using the typed api.

PR #697, opened 2020-06-24
add write sharing to simulator

kevodwyer:


src/peergos/server/simulation/AccessControl.java
```diff
@@ -94,9 +94,7 @@ private AccessControlUnit getAcu(FileSystem.Permission permission) {
                 case WRITE:
                     return writers.getAllowed(path);
                 case READ:
-                    List<String> allowed = readers.getAllowed(path);
-                    List<String> allowed2 = writers.getAllowed(path);
-                    return new ArrayList<>(Stream.of(allowed, allowed2).flatMap(e -> e.stream()).collect(Collectors.toSet()));
+                    return readers.getAllowed(path);
```
ianopolous, line 97: Can you explain this change? This was done because write access implies read access in cryptree.

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -1001,13 +1001,10 @@ public static int getNumberOfChunks(long size) {
...
         return current.withWriter(existingChild.owner(), existingChild.writer(), network)
-                .thenCompose(state -> (existingChild.isDirty() ?
+                .thenCompose(state ->
                         existingChild.clean(state, committer, network, crypto, parent)
-                                .thenCompose(pair -> pair.left.getChild(pair.right, filename, crypto.hasher, network)
-                                        .thenApply(cleanedChild -> cleanedChild.get())) :
-                        CompletableFuture.completedFuture(existingChild))
-                ).thenCompose(updatedChild -> updatedChild.overwriteSection(updatedChild.version, committer, fileData,
-                        inputStartIndex, endIndex, network, crypto, monitor));
+                                .thenCompose(pair -> pair.left.overwriteSection(pair.right, committer, fileData,
```
ianopolous, line 1006: What was wrong here?

src/peergos/server/simulation/AccessControl.java
kevodwyer: It still does. When verifying access permissions in verifySharingPermissions() it takes that into account.

src/peergos/shared/user/fs/FileWrapper.java
kevodwyer: previously clean() returned Pair<parent, Snapshot> (although it didn't do that in all cases). Now that an updated ref to the FW is returned, there is no reason to go from parent to child.

src/peergos/server/simulation/AccessControl.java
ianopolous: I'm not seeing where that happens. Where do you mean?

kevodwyer: Simulator.java line 506: case WRITE:  You will see that for write permission is first checks fs.read(path); and then fs.write(path, new byte[]{read[0]});

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -696,7 +711,7 @@ public SimulationSupplier(Map<Simulation, Double> probabilities, Random random)
         int seed = simulatorArgs.getInt("random-seed", 1);
         int nUsers = simulatorArgs.getInt("n-users", 3);
         int meanFileLength  = simulatorArgs.getInt("mean-file-length", 256);
-        boolean randomizeFriendNetwork = simulatorArgs.getBoolean("randomize-friend-network", true);
+        boolean randomizeFriendNetwork = simulatorArgs.getBoolean("randomize-friend-network", false);
```
ianopolous, line 714: what happens with this true? does it still verify?

kevodwyer: Currently if the friend network is randomised the code does not take the relationshsips into account when granting sharing access.

Issue #699, opened 2020-06-27
Unused code?


odisseus:
It seems that the class `peergos.server.storage.IPFS.IPFSObject` is not used anywhere, not even in tests.

PR #700, opened 2020-06-27
Deleted peergos.server.storage.IPFS.IPFSObject

odisseus:
This pull request is related to the issue #699.


ianopolous:
We don't currently use this in Peergos that's correct, but this is still part of the IPFS API.

odisseus:
Do those methods actually work with IPFS 0.5.x and 0.6.x?

ianopolous:
Yes

Issue #392, opened 2019-02-13
Fuzz Cryptree


ianopolous:
Use a host filesystem and a self contained permission model that mirrors our expected semantics to fuzz cryptree. 

By this I mean create a model and a cryptree filesystem and set of users, and do random operations and compare the two file systems and user access after each. 

Operations including:
* write new file
* mkdir
* modify file
* delete a file/dir
* copy a file/dir
* grant (R/W) access to a n users to a file/dir
* revoke (R/W) access to a user

kevodwyer:
actions to add: rm shared file, rm shared dir, rename shared file, rename shared dir

PR #705, opened 2020-06-30
uncomment shared file read and write

kevodwyer:


src/peergos/server/simulation/PeergosFileSystemImpl.java
```diff
@@ -50,6 +54,10 @@ private FileWrapper getDirectory(Path path) {
     @Override
     public void write(Path path, byte[] data, Consumer<Long> progressConsumer) {
         FileWrapper directory = getDirectory(path);
+        Optional<FileWrapper> existingFile = userContext.getByPath(path).join();
+        if (existingFile.isPresent() && existingFile.get().getFileProperties().isDirectory) {
+            return;
```
ianopolous, line 59: Couldn't you check it is writable?

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -515,16 +543,26 @@ private boolean verifySharingPermissions(String user, Path path) {
...
-                                fs.write(path, new byte[]{read[0]});
-                            } catch (Exception ex) {
-                                LOG.log(Level.SEVERE, "User " + sharee + " could not write  shared-path " + path + "!", ex);
-                                isVerified = false;
+                                fs.write(path, read == null ? null : new byte[]{read[0]});
+                            } catch (Throwable ex) {
+                                //reload user and try again
+                                fs = fileSystems.getTestFileSystem(sharee, true);
+                                try {
+                                    fs.write(path, read == null ? null : new byte[]{read[0]});
```
ianopolous, line 551: why is this necessary?

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -93,8 +93,8 @@ public void addUser(String user) {
     enum Simulation {
         READ_OWN_FILE,
         WRITE_OWN_FILE,
-        READ_SHARED_FILE,
-        WRITE_SHARED_FILE,
+        READ_SHARED_FILE_OR_DIRECTORY,
```
ianopolous, line 96: Do you think these should be two different things? reading a file is very different to reading a directory. The latter is doing a directory listing.

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -515,16 +543,26 @@ private boolean verifySharingPermissions(String user, Path path) {
                         if (isVerified) {
                             try {
                                 // can overwrite?
-                                fs.write(path, new byte[]{read[0]});
-                            } catch (Exception ex) {
-                                LOG.log(Level.SEVERE, "User " + sharee + " could not write  shared-path " + path + "!", ex);
-                                isVerified = false;
```
cboddy, line 521: Can you explain what the logic for this is?

src/peergos/server/tests/simulation/Simulator.java
cboddy: I think they should be different operations

src/peergos/server/tests/simulation/NativeFileSystemImpl.java
```diff
@@ -70,7 +70,9 @@ private void ensureCan(Path path, Permission permission, String user) {
     public byte[] read(Path path, BiConsumer<Long, Long> pc) {
         Path nativePath = virtualToNative(path);
         ensureCan(path, Permission.READ);
-
+        if (nativePath.toFile().isDirectory()) {
+           return null;
```
cboddy, line 74: This would be easier to reason about if it raised an exception. Why return null?

src/peergos/server/simulation/PeergosFileSystemImpl.java
```diff
@@ -42,6 +43,9 @@ private FileWrapper getDirectory(Path path) {
     public byte[] read(Path path, BiConsumer<Long, Long> progressConsumer) {
         FileWrapper wrapper = getPath(path);
         long size = wrapper.getFileProperties().size;
+        if (wrapper.getFileProperties().isDirectory) {
+            return null;
```
cboddy, line 47: This should raise

src/peergos/server/simulation/PeergosFileSystemImpl.java
kevodwyer: sure, also see answer below

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: That is the question. I don't know the answer at this stage. I am investigating. Making this change was sufficient for verify() to succeed.

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: ok, I was attempting to make the least amount of changes (no changes to interfaces). The thought being reading/writing actual file is only one step further past checking existence of file/directory.

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: working on it. The above solves the problem, but why the snapshot is out-of-date is not answered yet.

src/peergos/server/tests/simulation/Simulator.java
ianopolous: Maybe check it's not using a caching mutable pointers?

src/peergos/server/tests/simulation/NativeFileSystemImpl.java
kevodwyer: The method still confirms the file exists and has the right permissions. I didn't want to add read/write directory specific method to interface. But ok, message heard. I can do so.

src/peergos/server/simulation/PeergosFileSystemImpl.java
kevodwyer: same as above. I can revisit.

src/peergos/server/simulation/PeergosFileSystemImpl.java
kevodwyer: fixed

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: grant code did not take into account any previous call to grant

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: see new code

src/peergos/server/tests/simulation/NativeFileSystemImpl.java
kevodwyer: done now

src/peergos/server/simulation/PeergosFileSystemImpl.java
kevodwyer: separated file and directory cases now

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: removed - was due to simulator not taking into account existing sharing state

kevodwyer:
have updated PR to take into account review

kevodwyer:
I have been running it all afternoon. I have gone back to global verify after xyz operations else i run out of file descriptors on my machine.

kevodwyer:
I have been running it all afternoon. I have gone back to global verify after xyz operations else i run out of file descriptors on my machine.

kevodwyer:
PR  open until i sort out a rare bug

kevodwyer:
bug was not in simulator. this PR is ready to be reviewed

cboddy:
Sorry for missing the context but can you say what the objective of this change is, please?

kevodwyer:
Implement READ_SHARED_FILE, READ_SHARED_DIRECTORY, WRITE_SHARED_FILE, WRITE_SHARED_DIRECTORY

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -263,13 +270,18 @@ public Simulator(int opCount, Random random, long meanFileLength,
...
+    private void readFile(String user, Path path) {
         FileSystem testFileSystem = fileSystems.getTestFileSystem(user);
         FileSystem referenceFileSystem = fileSystems.getReferenceFileSystem(user);
+        referenceFileSystem.read(path);
+        testFileSystem.read(path);
+    }
 
-        byte[] refBytes = referenceFileSystem.read(path);
-        byte[] testBytes = testFileSystem.read(path);
-        return Arrays.equals(refBytes, testBytes);
```
cboddy, line 272: Why is  this not comparing the bytes anymore?

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -263,13 +270,18 @@ public Simulator(int opCount, Random random, long meanFileLength,
...
+    private void readDirectory(String user, Path path) {
```
cboddy, line 280: 1.  I think the concept of "reading" a  directory is a bit confusing  and doesn't directly map onto (is it read & execute?) - why introduce it?
2. What is this supposed to check?  Possible side effects of  `ls`-ing a  directory?

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -94,7 +94,9 @@ public void addUser(String user) {
         READ_OWN_FILE,
         WRITE_OWN_FILE,
         READ_SHARED_FILE,
+        READ_SHARED_DIRECTORY,
         WRITE_SHARED_FILE,
+        WRITE_SHARED_DIRECTORY,
```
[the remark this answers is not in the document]
ianopolous: read means you can read the contents. write means you can add files or subdirs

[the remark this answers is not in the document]
ianopolous: This is what unix does.

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -287,8 +299,10 @@ private void write(String user, Path path) {
         Path dirName = path.getParent();
         String fileName = path.getFileName().toString();
         Map<Path, List<String>> dirsToFiles = index.getDirToFiles(user);
-        dirsToFiles.get(dirName).add(fileName);
-
+        List<String> existingFiles = dirsToFiles.get(dirName);
+        if (! existingFiles.contains(fileName)) {
```
cboddy, line 303: this  makes me a bit nervous - should `existingFiles` be updated to a set instead?

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -301,6 +315,26 @@ private boolean grantPermission(String granter, String grantee, Path path, FileS
         FileSystem testFileSystem = fileSystems.getTestFileSystem(granter);
         FileSystem referenceFileSystem = fileSystems.getReferenceFileSystem(granter);
 
+        Set<String> testExistingWriters = new TreeSet<>(testFileSystem.getSharees(path, FileSystem.Permission.WRITE));
+        Set<String> refExistingWriters = new TreeSet<>(referenceFileSystem.getSharees(path, FileSystem.Permission.WRITE));
+        if (! testExistingWriters.toString().equals(refExistingWriters.toString())) {
```
cboddy, line 320: nit-pick: why convert to string when comparing?

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -376,35 +410,72 @@ private void run(final Simulation simulation, final String user) {
...
         switch (simulation) {
             case READ_OWN_FILE:
                 Path readPath = randomFilePathForUser.get();
                 log(user, simulation, readPath);
-                read(user, readPath);
+                readFile(user, readPath);
                 break;
             case READ_SHARED_FILE:
-                Optional<Path> sharedOpt = randomSharedPath.apply(otherUser.get(), user);
+                Optional<Path> sharedOpt = randomSharedFilePath.apply(user, otherUser.get());
```
cboddy, line 443: Formerly this would result in `user` reading a (random) file that another (random)  user had shared with `user` -  is that still the case?

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -376,35 +410,72 @@ private void run(final Simulation simulation, final String user) {
...
+                readDirectory(user, sharedDirPathToRead);
                 break;
             case WRITE_OWN_FILE:
                 Path path = getAvailableFilePath(user);
                 log(user, Simulation.WRITE_OWN_FILE, path);
                 write(user, path);
                 break;
             case WRITE_SHARED_FILE:
-                Optional<Path> sharedOpt2 = randomSharedPath.apply(otherUser.get(), user);
+                Optional<Path> sharedOpt2 = randomSharedFilePath.apply(user, otherUser.get());
```
cboddy, line 464: Same question as above

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -376,35 +410,72 @@ private void run(final Simulation simulation, final String user) {
...
                 if (! sharedOpt2.isPresent())
                     return;
                 Path  sharedPathToWrite = sharedOpt2.get();
                 log(user, Simulation.WRITE_SHARED_FILE, sharedPathToWrite);
                 write(user, sharedPathToWrite);
                 break;
+            case WRITE_SHARED_DIRECTORY:
```
cboddy, line 471: Can we call this `MKDIR_SHARED` instead, please?

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -461,8 +526,13 @@ public void run() {
             System.out.println("iOp=" + iOp);
             Simulation simulation = getNextSimulation.get();
             run(simulation);
+            /*
```
cboddy, line 529: commented  code to be removed

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -486,44 +558,79 @@ private boolean verifySharingPermissions(String user, Path path) {
                 LOG.info("User " + peergosFileSystem.user() + " path " + path + " has peergos-fs " + permission.name() + "ers " +
                         shareesInPeergos + " and local-fs " + permission.name() + "ers " + shareesInLocal);
                 isVerified = false;
+                break;
```
cboddy, line 561: Why break here?  Even if the permissions  are wrong we still want to check the contents  for  `user`/`path`.

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -486,44 +558,79 @@ private boolean verifySharingPermissions(String user, Path path) {
...
-                        try {
-                            //can read?
-                            read = fs.read(path);
-                        } catch (Exception ex) {
-                            LOG.log(Level.SEVERE, "User "+ sharee +" could not read shared-path "+ path +"!", ex);
-                            isVerified = false;
-                        }
-                        if (isVerified) {
+                PeergosFileSystemImpl fs = fileSystems.getTestFileSystem(sharee);
+                if (isDir) {
```
cboddy, line 568: I  think forcing reading and writing  for  directories and files into two ops makes this quite complicated.

Suggestion: simpligy by removing the `READ_SHARED_DIR` and `WRITE_SHARED_DIR` ops and add new `MKDIR_SHARED` and  `LS_SHARED` operations that will create or list the contents  of  a (random) shared directory,  or raise an error if  path is a file.

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -486,44 +558,79 @@ private boolean verifySharingPermissions(String user, Path path) {
...
+                            } catch (Exception ex) {
+                                LOG.log(Level.SEVERE, "User " + sharee + " could not read shared-path " + path + "!", ex);
+                                isVerified = false;
+                            }
+                            break;
+                        case WRITE:
                             try {
-                                // can overwrite?
-                                fs.write(path, new byte[]{read[0]});
+                                //can read?
```
cboddy, line 612: Isn't this no longer checking that the path can be written? It's only being read now.

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -486,44 +558,79 @@ private boolean verifySharingPermissions(String user, Path path) {
...
+                                read = fs.read(path);
                             } catch (Exception ex) {
-                                LOG.log(Level.SEVERE, "User " + sharee + " could not write  shared-path " + path + "!", ex);
+                                LOG.log(Level.SEVERE, "User " + sharee + " could not read shared-path " + path + "!", ex);
```
cboddy, line 615: Can you make the log message  explicit that this should be a _writable_ file that failed to be read?

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -486,44 +558,79 @@ private boolean verifySharingPermissions(String user, Path path) {
...
                                 isVerified = false;
                             }
-                        }
-                        break;
-                    default:
-                        throw new IllegalStateException();
+                            if (isVerified) {
```
cboddy, line 618: Why only do  this if  the permissions are  correct? Also this way it won't log the problem (that the path is  not writable).

src/peergos/server/tests/simulation/Simulator.java
ianopolous: Write is the terminology unix uses. We are just using mkdir to test it

src/peergos/server/tests/simulation/Simulator.java
```diff
@@ -535,13 +642,23 @@ private boolean verifySharingPermissions(String user, Path path) {
...
+        Stat stat = fileSystems.getReferenceFileSystem(user).stat(path);
+        boolean isDir = stat.fileProperties().isDirectory;
         try {
-            byte[] testData = peergosFileSystem.read(path);
-            byte[] refData = nativeFileSystem.read(path);
-            if (!Arrays.equals(testData, refData)) {
-                LOG.info("Path " + path + " has different contents between the file-systems");
-                return false;
+            if (isDir) {
+                Set<Path> refDirectoryListing = new TreeSet<>(nativeFileSystem.ls(path, true));
```
cboddy, line 649: I'm not sure this  is necessary -  by checking  each  _file_ in both filesystems the list of files in each directory is also implicitly checked.

What this will check is if there is some problem with the  result of the `ls` call in each directory - a view that should be the same as reading all the files we think are there. If we're going to check  `ls`  explicitly I think do  it separately - iterate through  all the directories and check the `ls` rather  than shoe-horning it in here.

ianopolous: You can be granted read access to things in a directory and not have full read access to the parent dir, this means you wouldn't be able to see siblings of the files. Similarly you can be granted write access to a file, but not even read to its parent. This means you can modify that file, but not rename it, nor ls the parent dir. This is what unix does.

cboddy: Yes, and this should still be checked separately.

src/peergos/server/tests/simulation/Simulator.java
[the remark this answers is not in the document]
cboddy: That depends on the execute bit  for that directory (in unix).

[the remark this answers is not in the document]
ianopolous: It is implemented as W && X, but that's not the semantic meaning, just an implementation detail: https://en.wikipedia.org/wiki/File_system_permissions

src/peergos/server/tests/simulation/Simulator.java
ianopolous: What about an empty directory? Just checking files won't catch that.

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: The return value was never checked previously anyway. The contents is now checked as part of verify()

src/peergos/server/tests/simulation/Simulator.java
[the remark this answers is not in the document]
cboddy: As discussed it's actually `rwx` right now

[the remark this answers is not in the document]
ianopolous: In unix speak yes, but semantically it's: R means you can list and read the children. W means you can add and rename children (and assumes you also have R)

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: 1. read == list. we use the word read for file so it seemed appropriate. we have write directory for make directory.
2. yes, the code can throw exceptions under a number of circumstances.

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: done

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: force of habit, fixed

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: 2 modes, verify after each op, or verify after all ops. the former useful when debugging, later is what it was originally

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: if the permissions are wrong there is a bug. When bug is fixed then the extra processing is done.

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: fixed

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: every time isVerfied is set to false a SEVERE log message is written. If that occurs, i expect the bug to be fixed and then the extra checks are done.

src/peergos/server/tests/simulation/Simulator.java
cboddy: The point is we still want to know if the contents match if the permissions are wrong

src/peergos/server/tests/simulation/Simulator.java
cboddy: Did you have a chance to check this?

kevodwyer: I'm still working through your comments. I get a fatal exception with the params in the original position. I hope to have time tomorrow to investigate further.

src/peergos/server/tests/simulation/Simulator.java
cboddy: Oh I see. Leo free to add a flag to do this too :)

You still need to check everything at the end in case of regressions that affect fonts/folders other than the path directly referenced in the simulation op.

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: Reversing the order of params like i did originally meant i could compare ref and peergos filesystems. 
As ref file system does not use sym links to model the shared dir structure i can't compare between the filesystems. That's fine i now just check to make sure the fie/dir exists in peergos dir.

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: answer above

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: done

src/peergos/server/tests/simulation/Simulator.java
kevodwyer: I'm reluctant to change this again. It does the job and i think it is clear. Can be enhanced in a subsequent PR.

kevodwyer:
@cboddy @ianopolous I think i have addressed all outstanding comments. Have run simulator again a number of times and all looks good.  There is a an item on the project board for further fuzzer work that can be done in another PR. https://github.com/Peergos/Peergos/issues/392

src/peergos/server/tests/simulation/Simulator.java
ianopolous: The fact that it is doing a mkdir and ls is just an implementation detail. Semantically it is testing whether we can read the directory, or write to the directory. And indeed we could change the implementation to for example delete the dir made. Instead of ls we could also check we have a read key in the Peergos case.

ianopolous:
If no objections by tomorrow, we should merge this.

PR #706, opened 2020-07-01
truncate bug

kevodwyer:


src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -768,7 +774,7 @@ public static int getNumberOfChunks(long size) {
         if (isWritable())
             return network.synchronizer.applyComplexUpdate(owner(), signingPair(), (current, committer) ->
                     uploadFileSection(current, committer, filename, fileData, isHidden, startIndex, endIndex,
-                            baseKey, overwriteExisting, network, crypto, monitor, firstChunkMapKey))
+                            baseKey, overwriteExisting, true, network, crypto, monitor, firstChunkMapKey))
```
ianopolous, line 777: I don't agree with this change. I'll push a fix.

PR #707, opened 2020-07-01
[print-test-errors]

cboddy:
This  script  parses the junit logs in  test.reports and prints any
testcase that had an error, including the stacktrace.

### DIFF
--- a/src/peergos/server/tests/CborObjects.java
+++ b/src/peergos/server/tests/CborObjects.java
@@ -18,7 +18,6 @@ public class CborObjects {
 
     @Test
     public void dosCborObject() throws Throwable {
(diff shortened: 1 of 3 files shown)


src/peergos/server/tests/CborObjects.java
```diff
@@ -18,6 +18,7 @@
 
     @Test
     public void dosCborObject() throws Throwable {
+        if(true)throw new RuntimeException();
```
cboddy, line 21: temporary?

ianopolous: yep, will revert once I check it works on github

PR #708, opened 2020-07-02
handle case where mp3 thumbnail cannot be extracted

kevodwyer:


kevodwyer:
> This still doesn't ensure the future on line 1602 is always completed. We need an exceptionally clause to propagate the error to fut, on line 1664.

done

PR #710, opened 2020-07-09
Server <==> user messages

ianopolous:
This implements a mechanism for a server operator to message users of that server. And for the users to reply, or indeed initiate conversations via feedback. 

This can be used for talking to our users, getting feedback, and surveys. There is a CLI for the server operator to add messages, or read replies from users. 

New messages are shown at login. 

See corresponding web-ui pr: https://github.com/Peergos/web-ui/pull/161


src/peergos/shared/user/ServerConversation.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public ServerConversation(List<ServerMessage> messages) {
+        this.messages = messages;
+        long latest = 0;
+        for (ServerMessage message : messages) {
+            if (message.sentEpochMillis > latest)
+                latest = message.sentEpochMillis;
+        }
+        ServerMessage lastMessage = messages.get(messages.size() - 1);
+        this.isDismissed = lastMessage.isDismissed || lastMessage.type == ServerMessage.Type.FromUser;
```
kevodwyer, line 22: The last part of this statement is surprising (but i can see why you did it).

ianopolous: Isn't it just the logic you asked for?

kevodwyer: Yes. It is the correct logic to filter for display. Using the logic to set a variable called isDismissed is what i question.

ianopolous: Oh sure. I can rename it. isHidden?

kevodwyer: IsDisplayable maybe?

ianopolous: done

src/peergos/server/messages/ServerMessageStore.java
```diff
@@ -0,0 +1,175 @@
...
+import java.util.*;
+import java.util.concurrent.*;
+import java.util.function.*;
+import java.util.logging.*;
+import java.util.stream.*;
+
+public class ServerMessageStore implements ServerMessager {
+    private static final Logger LOG = Logging.LOG();
+
+    private static final String SELECT = "SELECT id, type, sent, body, priorid, dismissed FROM messages WHERE username = ?;";
```
kevodwyer, line 23: SQL is good at sorting. If returned in chronological order from DB there is less work to do in Java code later.

ianopolous: I prefer to keep it simple, and not rely on the db for too much. This results is greater portability across databases.

PR #713, opened 2020-07-14
Fix/s3 error fallback

ianopolous:


src/peergos/shared/storage/RetryStorage.java
```diff
@@ -42,28 +41,30 @@ private int jitter(int minMilliseconds, int rangeMilliseconds) {
...
-                i-> {
-                    CompletableFuture<V> res = new CompletableFuture<>();
-                    f.get()
-                            .thenAccept(res::complete)
-                            .exceptionally(e ->  {
-                                if (i == 1) {
-                                    res.completeExceptionally(e);
-                                } else {
-                                    retryAfter(() -> recurse(i - 1, f)
+        CompletableFuture<V> res = new CompletableFuture<>();
```
kevodwyer, line 44: how about moving CF res up a level into runWithRetry and passing it through as a method param?

src/peergos/shared/storage/RetryStorage.java
```diff
@@ -42,28 +41,30 @@ private int jitter(int minMilliseconds, int rangeMilliseconds) {
...
+        try {
+            f.get()
+                    .thenAccept(res::complete)
+                    .exceptionally(e -> {
+                        if (retriesLeft == 1) {
+                            res.completeExceptionally(e);
+                        } else if (e instanceof HttpFileNotFoundException) {
```
kevodwyer, line 51: That's smart!

src/peergos/shared/storage/DirectS3BlockStore.java
```diff
@@ -211,7 +212,17 @@ public static Multihash keyToHash(String keyFileName) {
                             res[i] = new FragmentWithHash(new Fragment(identity.getHash()), Optional.empty());
                         }
                     return Arrays.asList(res);
+                }).thenAccept(allResults::complete)
+                .exceptionally(t -> {
+                    NetworkAccess.downloadFragments(hashes, this, monitor, spaceIncreaseFactor)
```
kevodwyer, line 217: Good to add a comment here to explain fallback

ianopolous: Is it not clear given it is the same as the call at the beginning of the method?

src/peergos/shared/storage/RetryStorage.java
ianopolous: I think it's cleaner to not have references being passed around.

src/peergos/shared/storage/DirectS3BlockStore.java
kevodwyer: not that bothered

src/peergos/shared/storage/RetryStorage.java
kevodwyer: ok

PR #717, opened 2020-07-18
Generalized GC

ianopolous:
This refactors the GC algorithm and makes sure that we can apply it to any datastore, as well as making it accessible to tests.

We use two, so far unused, ipfs calls block.rm and refs.local to achieve this for any ipfs instance 
(even non S3 backed ones). 

There is also a new social benchmark to test login speed with many friends, shares, etc.


src/peergos/server/storage/TransactionalIpfs.java
```diff
@@ -0,0 +1,78 @@
...
+                                                  PublicKeyHash writer,
+                                                  List<byte[]> signedHashes,
+                                                  List<byte[]> blocks,
+                                                  TransactionId tid) {
+        for (byte[] signedHash : signedHashes) {
+            Multihash hash = new Multihash(Multihash.Type.sha2_256, Arrays.copyOfRange(signedHash, signedHash.length - 32, signedHash.length));
+            Cid cid = new Cid(1, Cid.Codec.DagCbor, hash.type, hash.getHash());
+            transactions.addBlock(cid, tid, owner);
+        }
+        return target.put(owner, writer, signedHashes, blocks, tid);
```
kevodwyer, line 46: feels like a 2 stage commit

ianopolous: All we require is that it is in the transaction store before it is written.

PR #720, opened 2020-07-24
Rewrite shared-with cache

ianopolous:
This rewrites the shared-with cache to reside in a parallel directory tree. This then contributes 0 to login time, and O(1) to navigate to a directory and display which children are shared.

### DIFF
--- a/src/peergos/shared/user/SharedWithState.java
+++ b/src/peergos/shared/user/SharedWithState.java
@@ -31,6 +31,7 @@ public class SharedWithState implements Cborable {
         return writeShares;
     }
 
+    @JsMethod
     public FileSharedWithState get(String filename) {
         return new FileSharedWithState(
                 readShares.getOrDefault(filename, Collections.emptySet()),
--- a/src/peergos/shared/user/SharedWithState.java
+++ b/src/peergos/shared/user/SharedWithState.java
@@ -56,6 +56,8 @@ public class SharedWithState implements Cborable {
     }
 
     public SharedWithState add(SharedWithCache.Access access, String filename, Set<String> names) {
+        if (names.isEmpty())
+            return this;
         Map<String, Set<String>> newReads = new HashMap<>();
         for (Map.Entry<String, Set<String>> e : readShares.entrySet()) {
             newReads.put(e.getKey(), new HashSet<>(e.getValue()));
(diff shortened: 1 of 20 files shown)


src/peergos/shared/user/SharedWithState.java
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        newReads.remove(filename);
+        newWrites.remove(filename);
+
+        return new SharedWithState(newReads, newWrites);
+    }
+
+    @JsMethod
+    public boolean isShared(String filename) {
```
kevodwyer, line 112: SharedWithState is internal. Not sure why this is a jsmethod.

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -418,11 +418,6 @@ public boolean isLink() {
         return linkPointer.isPresent();
     }
 
-    @JsMethod
```
kevodwyer, line 421: What is the alternative?

ianopolous: UserContext.sharedWith(path), but I can add something that return a boolean

src/peergos/shared/user/SharedWithState.java
ianopolous: It's not internal. It's meant to be used by the UI directly. You retrieve it once for the current dir.

ianopolous: I've added a helped method UserContext.getDirSharingState for this

src/peergos/shared/user/fs/FileWrapper.java
ianopolous: Actually I removed the boolean method, see replies to other comment

src/peergos/shared/user/SharedWithState.java
kevodwyer: I think i'm missing something.
I can't see what class available to the UI returns an instance of SharedWithCache

kevodwyer: now up-to-date.

ianopolous:
Any more comments here @kevodwyer ?

PR #725, opened 2020-08-06
Use config file by default, and save any args to it.

ianopolous:
This will make it much easier to run other commands like the server-msg commands with matching args.

### DIFF
--- a/src/peergos/server/util/Args.java
+++ b/src/peergos/server/util/Args.java
@@ -219,6 +219,15 @@ public class Args {
         }
     }
 
+    /** Save all parameters to a config file if it doesn't exist already
+     *
+     */
+    public void saveToFileIfAbsent() {
+        Path config = fromPeergosDir(CONFIG_FILENAME, CONFIG_FILENAME);
+        if (! config.toFile().exists())
+            saveToFile(config);
+    }
+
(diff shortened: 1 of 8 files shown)


src/peergos/server/util/Args.java
```diff
@@ -199,7 +217,25 @@ public Path getPeergosDir() {
...
+
+    private void saveToFile(Path file) {
+        String text = new TreeMap<>(params).entrySet().stream()
+                .filter(e -> ! envOnly.containsKey(e.getKey()))
+                .map(e -> e.getKey() + " = " + e.getValue())
+                .collect(Collectors.joining("\n"));
+        try {
+            Files.write(file, text.getBytes(), StandardOpenOption.CREATE);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
kevodwyer, line 237: If file already exists, should it raise a runtime ex?

ianopolous: I considered that, but I want to be able to add a cmd arg (either updating an existing one, or adding a new one) and have it end up in the config file.

ianopolous: On balance you're probably right, and I should just edit the config file. I'll change to only write the file if it doesn't exist (not throw).

PR #735, opened 2020-08-31
Implement a social feed

ianopolous:
This is  a ~chronological feed of things that have been shared with you.

It's only strictly chronological for things from a particular friend. We batch process things from each friend (in the client) so things from different friends may not be chronologically sorted. 

This also fixes loading of cached caps from a friend which was always trying to read the wrong file!


src/peergos/shared/social/SocialFeed.java
```diff
@@ -0,0 +1,285 @@
...
+                updatedReadCapBytesTotal,
+                updateWriteCapBytesTotal);
+        currentCapBytesProcessed.put(friendName, updated);
+        feedSizeRecords += readCapsToAdd.size() + writeCapsToAdd.size();
+        List<SharedItem> forFeed = Stream.of(readCapsToAdd, writeCapsToAdd)
+                .flatMap(List::stream)
+                .map(c -> new SharedItem(c.cap, extractOwner(c.path), friendName, c.path))
+                .collect(Collectors.toList());
+        ByteArrayOutputStream bout = new ByteArrayOutputStream();
+        for (SharedItem item : forFeed) {
```
kevodwyer, line 160: why not a stream rather than a loop?

src/peergos/shared/user/fs/AsyncReader.java
```diff
@@ -78,6 +78,64 @@
...
+                        try {
+                            CborObject readObject = CborObject.read(in, bytesRead);
+                            if (toSkip > 0)
+                                toSkip--;
+                            else {
+                                objectsToRead--;
+                                accumulator.accept(fromCbor.apply(readObject));
+                            }
+                            localOffset += readObject.toByteArray().length;
+                        } catch (RuntimeException e) {
```
kevodwyer, line 127: need a description here to explain exception and consequence

ianopolous: This is just to handle objects that straddle the chunk boundary

src/peergos/shared/social/SocialFeed.java
ianopolous: I don't like try catches inside streams - a loop is easier to read then I think.

src/peergos/shared/social/SocialFeed.java
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     *
+     * @param index
+     * @return the byte offset and corresponding index of a prior object boundary, which ideally should be in the same chunk
+     */
+    private CompletableFuture<Pair<Long, Integer>> getPriorByteOffset(int index) {
+        return dataDir.getChild(FEED_INDEX, context.crypto.hasher, context.network)
+                .thenCompose(fopt -> {
+                    //TODO
```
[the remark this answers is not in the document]
ianopolous: Not in this PR, and not until we've proved it is necessary.

kevodwyer (approved):

kevodwyer (approved):

kevodwyer (approved):

PR #739, opened 2020-09-05
Generalize CHAMP to allow any cborable value

ianopolous:
This will allow us to choose whether we want to inline values, or use merkle links.


kevodwyer (approved):
Any performance hit?

ianopolous:
It's identical for existing champs, so no performance change there.

PR #740, opened 2020-09-06
Fix following state machine to allow going from unidirectional

ianopolous:
to bidirectional


ianopolous:
Fixes: https://github.com/Peergos/Peergos/issues/688

kevodwyer (approved):

PR #741, opened 2020-09-10
Implement an inode based filesystem.

ianopolous:
This is used for publishing capabilities, and in the future will be for caching caps shared with you.

The top level is a champ mapping inode+path element to dir. Each large directory ends up as a sub champ mapping name to InodeCap


kevodwyer (commented):
all good so far

kevodwyer (approved):

PR #747, opened 2020-09-24
Clean up the separation between Java and JS implementations

ianopolous:


src/peergos/shared/crypto/password/PasswordProtected.java
```diff
@@ -31,7 +30,7 @@ public static Cborable encryptWithPassword(byte[] cleartext,
                                                Salsa20Poly1305 provider,
                                                SafeRandom random) {
         List<Cborable> elements = new ArrayList<>();
-        byte[] saltBytes = random.randomBytes(TweetNaCl.SECRETBOX_KEY_BYTES);
+        byte[] saltBytes = random.randomBytes(32);
```
kevodwyer, line 33: use of constant

ianopolous: This has nothing to do with secretbox, and isn't used anywhere else.

kevodwyer (approved):

PR #749, opened 2020-10-08
get rid of unnecessary recalculation of mime type during upload

kevodwyer:


ianopolous (approved):

ianopolous:
How many times was it being called?

PR #750, opened 2020-10-16
16X speedup of general operations

ianopolous:
This modifies the cryptree data structure to include the names of the children of a directory with the links to them (all padded and encrypted) in a backwards compatible way. Existing directories are converted to the new format when they are modified. 

This makes dir.getChild(name) O(1) rather than O(# children in dir). This means login is O(1) and in a benchmark with 50ms blockstore latency and 80 files in the home dir login is reduced from 80s to 5s. getByPath benefits from the same speed up. It also flows through to sharing, and sending follow requests. Upload and download speed are unchanged. 

Most of the interesting changes are in CryptreeNode, ChildrenLinks, and the two new classes NamedRelativeCapability and NamedAbsoluteCapability.

### DIFF
--- a/src/peergos/shared/user/fs/FileWrapper.java
+++ b/src/peergos/shared/user/fs/FileWrapper.java
@@ -233,8 +233,8 @@ public class FileWrapper {
 
     public CompletableFuture<Boolean> hasChildWithName(Snapshot version, String name, Hasher hasher, NetworkAccess network) {
         ensureUnmodified();
-        return getChildren(version, hasher, network)
-                .thenApply(children -> children.stream().anyMatch(c -> c.props.name.equals(name)));
+        return getChild(version, name, hasher, network)
+                .thenApply(Optional::isPresent);
     }
 
     /**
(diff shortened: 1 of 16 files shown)


ianopolous:
I will also test this on existing accounts with NonWriteThroughNetwork to test backwards compatibility and upgradeability. 

All fuzzer runs seem to be green.

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -417,31 +395,34 @@ public SymmetricKey getParentKey() {
...
-
     @JsMethod
     public CompletableFuture<Optional<FileWrapper>> getChild(String name, Hasher hasher, NetworkAccess network) {
         return getChild(version, name, hasher, network);
     }
 
-    public CompletableFuture<Optional<FileWrapper>> getChild(Snapshot version, String name, Hasher hasher, NetworkAccess network) {
-        return getChildren(version, hasher, network)
-                .thenApply(children -> children.stream().filter(f -> f.getName().equals(name)).findAny());
+    public CompletableFuture<Optional<FileWrapper>> getChild(Snapshot version,
```
kevodwyer, line 403: I think we also need a getLatestChild() method in case of a concurrent modification by another

ianopolous: I'm not sure what you mean. You can always call getUpdated on the child if desired?

ianopolous:
I've run tests on legacy data using a delaying non write through view and login has dropped from 15s to 2.4s!

kevodwyer (approved):

PR #751, opened 2020-10-16
todo board

kevodwyer:


src/peergos/shared/user/FileUtil.java
```diff
@@ -0,0 +1,53 @@
...
+
+    public static Path canonicalise(Path p) {
+        return p.isAbsolute() ? p : Paths.get("/").resolve(p);
+    }
+
+    public static Path toRelative(Path p) {
+        return p.isAbsolute() ? Paths.get(p.toString().substring(1)) : p;
+    }
+
+    public static CompletableFuture<FileWrapper> getOrMkdir(FileWrapper parent, String dirName, Crypto crypto, NetworkAccess network) {
```
ianopolous, line 34: getOrMkdir variants can be removed because it exists on FileWrapper now

src/peergos/shared/user/FileUtil.java
```diff
@@ -0,0 +1,53 @@
...
+import java.nio.file.Paths;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public class FileUtil {
+
+
+    public static String getFilename(Path p) {
```
ianopolous, line 18: can just use path.getFileName().toString()

src/peergos/shared/user/TodoBoard.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.*;
+import java.util.stream.Collectors;
+
+@JsType
+public class TodoBoard implements Cborable {
+
+    private static final String VERSION_1 = "1";
+    private final List<TodoList> todoLists;
+    private final String name;
+    private final LocalDateTime timestamp;
```
ianopolous, line 18: timestamp of what?

src/peergos/shared/user/fs/MimeTypes.java
```diff
@@ -121,6 +123,10 @@ public static final String calculateMimeType(byte[] start, String filename) {
                 return "text/calendar";
             return "text/plain";
         }
+
+        if (filename.endsWith(UserContext.App.Todo.TODO_FILE_EXTENSION))
```
ianopolous, line 127: I'd like to figure out a way to do this without relying on the extension, by making the cbor have a fixed header.
Probably having cbor = list(mimetype, map=data) will do it.

src/peergos/shared/user/UserContext.java
```diff
@@ -118,6 +119,179 @@ public TransactionService getTransactionService() {
         );
     }
 
+    @JsMethod
+    public App.Todo getTodoApp() {
```
ianopolous, line 123: All this being built in to Peergos concerns me. A large part of the point of the todo list is to dog food how we expect others to build apps that are isolated from our code. That means minimal changes to things like UserContext, and certainly nothing this app-specific.

src/peergos/shared/user/FileUtil.java
kevodwyer: method removed

src/peergos/shared/user/TodoBoard.java
kevodwyer: The TodoBoard. This is for change detection. If A shared a todo board with write access to B. A or B can smash the changes made by the other.  This highlighted a bug in overwriteFile() whereby the file properties were not been updated correctly

src/peergos/shared/user/FileUtil.java
kevodwyer: done

src/peergos/shared/user/TodoBoard.java
ianopolous: We should prevent overwriting changes the same way we do with the filewriter, using the CAS on file version.

ianopolous:
Can we do an actual rebase? There should be no resulting merge commits in this branch.

src/peergos/shared/user/fs/MimeTypes.java
kevodwyer: done - but not sure about this

src/peergos/shared/user/UserContext.java
kevodwyer: This is version 1. I would like to see the above in a separate PR.

ianopolous: Fine to do it this way first, but I don't think we should merge or deploy it like this.

src/peergos/shared/user/fs/MimeTypes.java
```diff
@@ -150,13 +156,30 @@ public static final String calculateMimeType(byte[] start, String filename) {
...
+            if (cborMimeType.get().equals(UserContext.App.Todo.TODO_MIME_TYPE) && filename.endsWith(UserContext.App.Todo.TODO_FILE_EXTENSION)) {
+                return UserContext.App.Todo.TODO_MIME_TYPE;
+            }
+        }
         return "application/octet-stream";
     }
 
+    private static Optional<String> extractCborMimeType(byte[] data) {
+        try {
+            CborDecoder decoder = new CborDecoder(new ByteArrayInputStream(data));
```
ianopolous, line 170: we should be able to do this without any explicit dependency on cbor code, just look at the raw bytes.

src/peergos/shared/user/fs/MimeTypes.java
```diff
@@ -43,7 +49,7 @@
     final static int[] WOFF = new int[]{'w','O','F','F'};
     final static int[] WOFF2 = new int[]{'w','O','F','2'};
 
-    final static int HEADER_BYTES_TO_IDENTIFY_MIME_TYPE = 28;
+    final static int HEADER_BYTES_TO_IDENTIFY_MIME_TYPE = 40;
```
ianopolous, line 52: ideally we wouldn't increase this. I think we can get away with using an integer, and then having a lookup from integer to mimetype in code. cbor is a binary type after all.

ianopolous:
Blockers to merging:
1. Remove share and delete buttons from todo app
2. Let user choose where to save todo file (in normal folders)
3. When user opens a todo file open it in the app. 
4. No social feed integration.
5. Atomic saving of the file, just like the text editor

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -444,6 +444,12 @@ public SymmetricKey getParentKey() {
                 .thenApply(children -> children.stream().filter(f -> f.getName().equals(name)).findAny());
     }
 
+    public CompletableFuture<Optional<FileWrapper>> getUpdatedChild(String name, Hasher hasher, NetworkAccess network) {
```
ianopolous, line 447: Can you explain why we need this?

kevodwyer: Was part of optimistic locking check. No longer required

kevodwyer:
I believe all the blockers have now been addressed

ianopolous (approved):

src/peergos/shared/user/fs/FileWrapper.java
ianopolous: Can we remove it if it's no longer needed?

src/peergos/shared/user/SharedItemCacheFile.java
```diff
@@ -0,0 +1,33 @@
+package peergos.shared.user;
+
+import peergos.shared.cbor.CborObject;
+import peergos.shared.cbor.Cborable;
+import peergos.shared.social.SharedItem;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class SharedItemCacheFile implements Cborable {
```
ianopolous, line 10: can this be deleted now?

ianopolous (approved):

PR #752, opened 2020-10-16
calendar

kevodwyer:
Implement a calendar.
Folder structure:
username/.apps/calendar/data/year/month
This is in anticipation that in the future it may be possible to have multipe calendars

Files are stored as .ics
Fields used:
UID, SUMMARY, DESCRIPTION, LOCATION DTSTART, DTEND, X-CATEGORY, STATUS
Duration is handled and converted to dtstart, dtend
UTC is used

A new class SharedItemCache is introduced.
This is used to store Caps shared with us. The cbor formatted file includes an index.

The cache file is stored in username/.apps/calendar


src/peergos/server/tests/UserTests.java
```diff
@@ -1155,12 +1155,14 @@ public void todoTest() throws Exception {
         FileWrapper userRoot = context.getUserRoot().join();
         final String TODO_FILE_EXTENSION = ".todo";
         FileWrapper updatedRoot = userRoot.uploadOrReplaceFile(todoBoardName + TODO_FILE_EXTENSION, new AsyncReader.ArrayBacked(data), data.length,
-                context.network, context.crypto, l -> {}, context.crypto.random.randomBytes(32)).get();
+                context.network, context.crypto, l -> {
```
ianopolous, line 1158: can we lose this formatting change?

src/peergos/server/util/PeergosNetworkUtils.java
```diff
@@ -13,11 +13,13 @@
 import peergos.shared.user.fs.FileWrapper;
 import peergos.shared.user.fs.cryptree.*;
 import peergos.shared.util.ArrayOps;
+import peergos.shared.util.Pair;
```
ianopolous: presumably these new imports aren't used?

src/peergos/server/tests/UserTests.java
kevodwyer: done

src/peergos/server/util/PeergosNetworkUtils.java
kevodwyer: fixed

src/peergos/shared/user/SharedItemCache.java
```diff
@@ -0,0 +1,48 @@
...
+import peergos.shared.cbor.CborObject;
+import peergos.shared.cbor.Cborable;
+import peergos.shared.social.SharedItem;
+
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class SharedItemCache implements Cborable {
+
+    private final List<SharedItem> items;
```
ianopolous, line 12: We need to think about this because apps should never have access to any keys I believe, and ideally paths if possible.

ianopolous:
We need the app internal storage to be in a sub dir. 
So $username/.apps/calendar/data
Then the paths the app deals with internally are all relative to that dir. It allows us to store for example things like what permissions have been granted in another file in /calendar dir

kevodwyer:
> We need the app internal storage to be in a sub dir.
> So $username/.apps/calendar/data
> Then the paths the app deals with internally are all relative to that dir. It allows us to store for example things like what permissions have been granted in another file in /calendar dir

I'm not sure i follow. You mean all files including sharedWithUs.cbor should be inside data?

src/peergos/shared/user/App.java
```diff
@@ -118,6 +119,18 @@ private Path fullPath(Path path) {
         });
     }
 
+    @JsMethod
+    public CompletableFuture<List<String>> dirInternal(Path relativePath) {
+        Path path = fullPath(relativePath);
+        return ctx.getByPath(path).thenCompose(dirOpt -> {
```
ianopolous, line 125: there is a getChildren call on usercontext

kevodwyer: I tried that, but the version on user context blows up if the path does not exist (at least it does from gwt js). The above returns an empty list.

ianopolous: ok let's fix it to cover that case

src/peergos/shared/util/Futures.java
```diff
@@ -9,6 +9,7 @@
 
 public class Futures {
 
+    @JsMethod
     public static final <T> CompletableFuture<T> of(T val) {
```
ianopolous, line 13: I can't find where this is used in JS?

kevodwyer: The calling code in JS was removed during a refactor. The .of() method is really useful though. I think it should be accesisble to Javascript. It would tidy up some code in gwt.js for a start.

src/peergos/shared/user/SharedItemCache.java
kevodwyer: removed

src/peergos/shared/util/Futures.java
ianopolous: gwt.js is a precursor to peergoslib so it can't depend on it.

ianopolous: Every JS annotation makes it harder to refactor safely, so I generally try to minimise them.

ianopolous (approved):

PR #753, opened 2020-10-19
handle case where an item was unshared

kevodwyer:


src/peergos/shared/user/UserContext.java
```diff
@@ -1650,13 +1650,22 @@ private static void ensureAllowedToShare(FileWrapper file, String ourname, boole
...
-        return Futures.combineAllInOrder(pointers.stream()
-                .map(s -> network.getFile(s.cap, s.owner))
-                .collect(Collectors.toList()))
-                .thenApply(res -> res.stream()
-                        .flatMap(Optional::stream)
-                        .collect(Collectors.toList()));
+    public CompletableFuture<List<Pair<SharedItem,FileWrapper>>> getFiles(List<SharedItem> pointers) {
+        List<CompletableFuture<Optional<?>>> futures = pointers.stream().map(s -> network.getFile(s)).collect(Collectors.toList());
+        CompletableFuture<List<Pair<SharedItem, FileWrapper>>> identity = CompletableFuture.completedFuture(Collections.emptyList());
+        return futures.stream().reduce(identity,
```
ianopolous, line 1656: we need to use combineAllInOrder to preserver order.

src/peergos/shared/NetworkAccess.java
```diff
@@ -340,6 +340,16 @@ private static NetworkAccess build(ContentAddressedStorage dht,
                 });
     }
 
+    public CompletableFuture<Optional<?>> getFile(SharedItem item) {
```
ianopolous, line 343: no need for this method, just use the one we were using

src/peergos/shared/user/UserContext.java
ianopolous: you can change the original code to call.thenapply in the map to make a pair

src/peergos/shared/user/UserContext.java
```diff
@@ -1650,13 +1650,22 @@ private static void ensureAllowedToShare(FileWrapper file, String ourname, boole
...
+                (a, b) -> b.thenCompose(opt ->
+                        a.thenApply(set -> {
+                            ArrayList<Pair<SharedItem, FileWrapper>> combined = new ArrayList<>(set.size() + 1);
```
ianopolous, line 1659: this would make the algorithm O(n^2) in the size of pointers

src/peergos/shared/user/UserContext.java
kevodwyer: This IS the code from combineAllInOrder. I moved it inline because i couldn't add the Pair<>()

src/peergos/shared/NetworkAccess.java
kevodwyer: But i need the item in the Pair. I'm sure it culd be done another way, but i need sleep.

src/peergos/shared/user/UserContext.java
kevodwyer: That means combineAllInOrder suffers from this also.

src/peergos/shared/user/UserContext.java
ianopolous: This very different from the existing code. E.g. the existing doesn't explicitly allocate a list.

src/peergos/shared/user/UserContext.java
ianopolous: No it doesn't. The n squared comes from creating a new list at each step. First it has 1 element, then 2 then 3 ... 1+2+3...+n = n(n+1)/2

PR #755, opened 2020-10-23
Private Profile

ianopolous:
Implement the Java side of a private profile, which is mainly just a convention of paths to profile elements. 

the only non trivial part is publishing a web root, which has 2 steps.


kevodwyer (approved):
GWT annotations can come later

PR #757, opened 2020-10-24
Don't use the direct S3 blockstore during signup.

ianopolous:
Otherwise users which block 3rd party domains will end up with an unrecoverable signup.

The important changes are in UserContext and NetworkAccess. 

Container implementations of ContentAddressedStorage should make a new version of themselves with a recursive call on their target.


ianopolous:
I've confirmed it works on test - I can log in after a failed sign up.

kevodwyer (approved):

PR #759, opened 2020-10-27
Put a 2 layer cache in the gateway.

ianopolous:
This reduces typical response time from 100-700ms down to 7-50ms

### DIFF
--- a/src/peergos/server/net/GatewayHandler.java
+++ b/src/peergos/server/net/GatewayHandler.java
@@ -16,15 +16,46 @@ import java.util.logging.*;
 public class GatewayHandler implements HttpHandler {
 	private static final Logger LOG = Logging.LOG();
 	private static final boolean LOGGING = true;
+	private static final int MAX_ASSET_SIZE_CACHE = 200*1024;
 
     private final String domainSuffix;
     private final NetworkAccess network;
     private final Crypto crypto;
+    private final LRUCache<String, FileWrapper> webRootCache;
+    private final LRUCache<String, Asset> assetCache;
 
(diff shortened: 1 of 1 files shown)


src/peergos/server/net/GatewayHandler.java
```diff
@@ -16,15 +16,46 @@
 public class GatewayHandler implements HttpHandler {
 	private static final Logger LOG = Logging.LOG();
 	private static final boolean LOGGING = true;
+	private static final int MAX_ASSET_SIZE_CACHE = 200*1024;
 
     private final String domainSuffix;
     private final NetworkAccess network;
     private final Crypto crypto;
+    private final LRUCache<String, FileWrapper> webRootCache;
```
kevodwyer, line 24: parameterise?

src/peergos/server/net/GatewayHandler.java
```diff
@@ -16,15 +16,46 @@
...
+    private final LRUCache<String, Asset> assetCache;
 
     public GatewayHandler(String domainSuffix, Crypto crypto, NetworkAccess network) {
         this.domainSuffix = domainSuffix;
         this.crypto = crypto;
         this.network = network;
+        this.webRootCache = new LRUCache<>(1000);
```
kevodwyer, line 31: paramaterise size?

ianopolous: It's not random, it's chosen based on the maximum amount of ram I want the cache to use.

ianopolous: And the number of users we have

kevodwyer (approved):

PR #762, opened 2020-10-29
add calendar and todo mimetype handling

kevodwyer:


src/peergos/shared/user/fs/MimeTypes.java
```diff
@@ -142,9 +142,10 @@ public static final String calculateMimeType(byte[] start, String filename) {
         if (equalArrays(start, CBOR_PEERGOS_TODO))
             return "application/vnd.peergos-todo";
 
+        if (filename.endsWith(".ics") && equalArrays(start, ICS))
```
ianopolous, line 145: is the file not a text file?

kevodwyer: Not on Mac at least. The ascii test fails.

ianopolous: Can we figure out why? icla is meant to be UTF8 by default, and I'd like to understand what we're seeing outside ascii

kevodwyer: Turns out ascii test does succeed. My issue was that when i was testing i didn't take into account that the first file uploaded is actually the transaction file and not the target file.

ianopolous (approved):

PR #766, opened 2020-11-17
user profile

kevodwyer:


src/peergos/shared/user/ProfilePaths.java
```diff
@@ -43,66 +44,92 @@
...
     }
 
+    @JsMethod
+    public static CompletableFuture<Optional<String>> getProfilePhoto(String user, UserContext viewer) {
+        return getAndParse(Paths.get(user).resolve(PHOTO), String::new, viewer);
+    }
+
+    @JsMethod
+    public static CompletableFuture<Boolean> setProfilePhoto(UserContext user, String base64Str) {
+        return serializeAndSet(PHOTO, base64Str, String::getBytes, user);
```
ianopolous, line 64: why are we serializing as a string?

kevodwyer: Currently for file thumbail we pass a String back to the GWT java code. We then use:
byte[] bytesOfData = Base64.getDecoder().decode(base64Str);
In the other direction in GWT java code we use:
            String base64Data = Base64.getEncoder().encodeToString(thumbnail.get());
            return "data:image/png;base64," + base64Data;
By storing the profile thumbnail as String i avoid needing to do both. Happy to change to Optional<byte> if you want that instead.

ianopolous: I think if we're storing an image let's store it as a normal binary. If JS wants base64 it can use btoa or atob.

kevodwyer: done

ianopolous (approved):

PR #768, opened 2020-11-23
sql quotas

ianopolous:
Move quotas from a file to sql, and allow importing from file

### DIFF
--- a/src/peergos/server/space/JdbcQuotas.java
+++ b/src/peergos/server/space/JdbcQuotas.java
@@ -16,6 +16,7 @@ public class JdbcQuotas {
     private static final String SET_QUOTA = "UPDATE freequotas SET quota = ? WHERE name = ?;";
     private static final String GET_QUOTA = "SELECT quota FROM freequotas WHERE name = ?;";
     private static final String GET_ALL_QUOTAS = "SELECT name, quota FROM freequotas;";
+    private static final String COUNT_USERS = "SELECT COUNT (name) FROM freequotas;";
     private static final String REMOVE_USER = "DELETE FROM freequotas WHERE name = ?;";
 
     private final SqlSupplier commands;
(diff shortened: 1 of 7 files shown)


src/peergos/server/space/JdbcQuotas.java
```diff
@@ -0,0 +1,149 @@
...
+        try (Connection conn = getConnection()) {
+            commands.createTable(commands.createQuotasTableCommand(), conn);
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public void setQuota(String username, long quota) {
+        try (Connection conn = getConnection();
+             PreparedStatement createuser = conn.prepareStatement(commands.insertOrIgnoreCommand("INSERT ", "INTO freequotas (name, quota) VALUES(?, ?)"));
```
kevodwyer, line 55: extract sql string to constant as per other queries

src/peergos/server/space/JdbcQuotas.java
```diff
@@ -0,0 +1,149 @@
...
+            select.setString(1, username);
+            ResultSet rs = select.executeQuery();
+            return rs.next();
+        } catch (SQLException sqe) {
+            LOG.log(Level.WARNING, sqe.getMessage(), sqe);
+            throw new IllegalStateException(sqe);
+        }
+    }
+
+    public int numberOfUsers() {
```
kevodwyer, line 125: select count() is the general way to do that in SQL

ianopolous: done

src/peergos/server/space/JdbcQuotas.java
ianopolous: It can't be a static constant because it's generated dynamically.

kevodwyer: I meant "INTO freequotas (name, quota) VALUES(?, ?)"

ianopolous: But it doesn't make any sense on it's own

kevodwyer (approved):

PR #778, opened 2020-12-18
Welcome CLI message and version

ianopolous:


kevodwyer (approved):
Nice
How about having the version also in the web-ui on the log-in page somewhere?
(Useful to know that the server you are connecting to is running the latest).
github hash is also useful sometimes

ianopolous:
The version is already accessible to the ui I believe via the InstanceAdmin on NetworkAccess. And yep, long term we definitely want to prompt the user to refresh if their version is out of date.

PR #783, opened 2021-01-12
Implement private sender side Groups

ianopolous:
This gives an O(1) way to share something with an arbitrarily large group of people.
You can define your own groups.

Members of a group cannot see:
 1) other members in the group
 2) how many members there are
 3) the name of the group

 The most important groups are two predefined ones for friends and followers

When a user is removed from a group we need to rotate keys for everything ever shared with that group. This is clearly expensive, but also a rare operation in normal usage.

### DIFF
--- a/src/peergos/shared/user/Groups.java
+++ b/src/peergos/shared/user/Groups.java
@@ -1,0 +1,52 @@
+package peergos.shared.user;
+
+import peergos.shared.cbor.*;
+import peergos.shared.crypto.random.*;
+import peergos.shared.util.*;
+
+import java.util.*;
+import java.util.function.*;
+
+public class Groups implements Cborable {
+    public final Map<String, String> uidToGroupName;
+
+    public Groups(Map<String, String> uidToGroupName) {
+        this.uidToGroupName = uidToGroupName;
+    }
+
+    public static Groups generate(SafeRandom r) {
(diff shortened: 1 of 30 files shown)


kevodwyer (approved):

src/peergos/shared/user/Groups.java
```diff
@@ -0,0 +1,52 @@
...
+        Map<String, String> uidToNames = new TreeMap<>();
+        uidToNames.put(generateUid(r), SocialState.FRIENDS_GROUP_NAME);
+        uidToNames.put(generateUid(r), SocialState.FOLLOWERS_GROUP_NAME);
+        return new Groups(uidToNames);
+    }
+
+    /* Generate a uid that cannot clash with a username, but which is a valid filename
+     */
+    public static String generateUid(SafeRandom r) {
+        return "." + ArrayOps.bytesToHex(r.randomBytes(32));
```
kevodwyer, line 27: maybe also add a prefix lilke "." + "G_" + random. In case we need to add another set of files for some other purpose in the future

kevodwyer (approved):
This takes functionality to a new level!

src/peergos/shared/user/Groups.java
ianopolous: if we need to add other we can do that safely without changing groups. e.g. with a non hex character after the .

PR #787, opened 2021-01-26
Revisit App to support sharing

kevodwyer:
While adding whole calendar sharing i came across the need to support:
- CRUD actions for items shared with user (ie friend's calendar items)
- secret links (no username on context)

The change is backards compatible


src/peergos/shared/user/App.java
```diff
@@ -47,9 +47,11 @@ private App(UserContext ctx, Path appDataDirectory) {
 
     @JsMethod
     public static CompletableFuture<App> init(UserContext ctx, String appName) {
-        App app = new App(ctx, Paths.get(ctx.username, APPS_DIR_NAME, appName, DATA_DIR_NAME));
+        Path path = Paths.get(ctx.username == null ? "<guest>" : ctx.username, APPS_DIR_NAME, appName, DATA_DIR_NAME);
```
ianopolous, line 50: Making up a dir doesn't sit right with me. Maybe App doesn't need it? There is also UserContext.getEntryPath

src/peergos/shared/user/app/StoreLocalAppData.java
```diff
@@ -9,11 +9,11 @@
  */
 public interface StoreLocalAppData {
 
-    CompletableFuture<List<String>> dirInternal(Path relativePath);
+    CompletableFuture<List<String>> dirInternal(Path relativePath, String username);
```
ianopolous, line 12: This api doesn't make sense to me. Do we need a new ReadLocalAppData which this extends for secret links?

ianopolous: I think I understand this now. But I need to think through the implications.

src/peergos/shared/user/App.java
```diff
@@ -23,41 +21,59 @@
...
+                .thenCompose(root -> root.getOrMkdirs(path, ctx.network, true, ctx.crypto))
                 .thenApply(appDir -> app);
     }
 
-    private Path fullPath(Path path) {
+    private void validatePath(Path path) {
         String pathAsString = path.toString().trim();
-        Path relativePath = pathAsString.startsWith("/") ? Paths.get(pathAsString.substring(1)) : Paths.get(pathAsString);
-        return appDataDirectory.resolve(relativePath);
+        if (pathAsString.contains("..")) {
```
ianopolous, line 59: what if there is a filename like a..b
need to first split on /

src/peergos/shared/user/App.java
```diff
@@ -23,52 +21,59 @@
...
         this.ctx = ctx;
-        this.appDataDirectory = appDataDirectory;
+        validatePath(appDataDirectory);
+        this.appDataDirectoryWithoutUser = appDataDirectory;
     }
 
     @JsMethod
     public static CompletableFuture<App> init(UserContext ctx, String appName) {
-        Path path = Paths.get(ctx.username == null ? "<guest>" : ctx.username, APPS_DIR_NAME, appName, DATA_DIR_NAME);
+        Path path = Paths.get(APPS_DIR_NAME, appName, DATA_DIR_NAME);
```
ianopolous, line 49: can we name it appDataDir rather than path?

ianopolous (approved):

PR #801, opened 2021-03-08
update progress when decrypting inline data

kevodwyer:


src/peergos/shared/crypto/FragmentedPaddedCipherText.java
```diff
@@ -132,9 +132,13 @@ public static FragmentedPaddedCipherText fromCbor(Cborable cbor) {
...
                                                   ProgressConsumer<Long> monitor) {
         if (inlinedCipherText.isPresent()) {
-            if (header.isPresent())
-                return Futures.of(new CipherText(nonce, ArrayOps.concat(header.get(), inlinedCipherText.get())).decrypt(from, fromCbor));
-            return Futures.of(new CipherText(nonce, inlinedCipherText.get()).decrypt(from, fromCbor));
+            T cipherText = header.isPresent() ?
+                    new CipherText(nonce, ArrayOps.concat(header.get(), inlinedCipherText.get())).decrypt(from, fromCbor)
+                    : new CipherText(nonce, inlinedCipherText.get()).decrypt(from, fromCbor);
+            if (cipherText instanceof byte[]) {
+                monitor.accept((long) ((byte[]) (cipherText)).length);
```
ianopolous, line 139: Can we do this without casts and instanceof?

ianopolous: It's probably okay to just use the length of the ciphertext here. As we know it will complete the download regardless.

kevodwyer: > Can we do this without casts and instanceof?

No. Could be of type CryptreeNode.ChildrenLinks

kevodwyer: > It's probably okay to just use the length of the ciphertext here. As we know it will complete the download regardless.

I looked at the length of various structures. Uness i'm looking at the wrong one, the length does not match ((byte[]) (cipherText)).length

ianopolous: The ciphertext is strictly longer yes. A better solution would be to add a decrypt method that takes the progress counter.

src/peergos/shared/crypto/FragmentedPaddedCipherText.java
```diff
@@ -133,11 +133,11 @@ public static FragmentedPaddedCipherText fromCbor(Cborable cbor) {
...
         if (inlinedCipherText.isPresent()) {
             if (header.isPresent())
-                return Futures.of(new CipherText(nonce, ArrayOps.concat(header.get(), inlinedCipherText.get())).decrypt(from, fromCbor));
-            return Futures.of(new CipherText(nonce, inlinedCipherText.get()).decrypt(from, fromCbor));
+                return Futures.of(new CipherText(nonce, ArrayOps.concat(header.get(), inlinedCipherText.get())).decrypt(from, fromCbor, monitor));
+            return Futures.of(new CipherText(nonce, inlinedCipherText.get()).decrypt(from, fromCbor, monitor));
         }
         return network.dhtClient.downloadFragments(cipherTextFragments, monitor, 1.0)
-                .thenApply(fargs -> new CipherText(nonce, recombine(header, fargs)).decrypt(from, fromCbor));
+                .thenApply(fargs -> new CipherText(nonce, recombine(header, fargs)).decrypt(from, fromCbor, monitor));
```
ianopolous, line 140: won't this double count?

kevodwyer: fixed

src/peergos/shared/crypto/CipherText.java
```diff
@@ -43,4 +44,13 @@ public static CipherText fromCbor(Cborable cbor) {
         byte[] secret = from.decrypt(cipherText, nonce);
         return fromCbor.apply(CborObject.fromByteArray(secret));
     }
+
+    public <T> T decrypt(SymmetricKey from, Function<CborObject, T> fromCbor, ProgressConsumer<Long> monitor) {
+        byte[] secret = from.decrypt(cipherText, nonce);
+        T res = fromCbor.apply(CborObject.fromByteArray(secret));
+        if (res instanceof byte[]) {
```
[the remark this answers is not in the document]
kevodwyer: updated as requested

ianopolous (approved):

Issue #674, opened 2020-05-05
If signup fails halfway (claimed username but not written filesystem). Allow a second attempt to complete.


ianopolous:

ianopolous:
This *should* be solved by https://github.com/Peergos/Peergos/issues/486

ianopolous:
This is actually a big problem because anyone blocking 3rd party domains by default, e.g. with umatrix, will be guaranteed to hit this. The result is the username is taken, but the filesystem is not committed, and cannot be rectified.

ianopolous:
Another solution would be to store the username + public salt temporarily in the browser local storage to allow an 2nd attempt to succeed. But really we should make the whole thing a single call with a single payload.

ianopolous:
A short term fix for the domain issue is https://github.com/Peergos/Peergos/pull/757

PR #802, opened 2021-03-08
Implement an atomic signup call

ianopolous:
This just runs the signup locally, and then sends all the operations
to the server in a single call which the server then replays.


ianopolous:
Closes: https://github.com/Peergos/Peergos/issues/674

kevodwyer (approved):

PR #809, opened 2021-03-23
Hash media locally that are uploaded to the social feed

ianopolous:


src/peergos/shared/social/SocialFeed.java
```diff
@@ -95,7 +95,7 @@ public SocialFeed(FileWrapper dataDir,
                 .thenCompose(p -> p.right.uploadAndReturnFile(uuid, media, length, false,
                         network, crypto)
                         .thenCompose(f -> f.getInputStream(f.version.get(f.writer()).props, network, crypto, monitor)
```
kevodwyer, line 97: is getInputStream required?

ianopolous: good catch!

PR #810, opened 2021-03-24
catch js storage exception

kevodwyer:
related UI PR: https://github.com/Peergos/web-ui/pull/237


ianopolous (approved):
Could we make it safer with a new Exception class?

kevodwyer:
> Could we make it safer with a new Exception class? Similar to HttpFileNotFond exception.
A 400 error could be caused by a number of things. Not as clean cut as the case with HttpFileNotFound

ianopolous:
I'm not saying all 400's become this new exception. Just check the trailer?

ianopolous (approved):

PR #818, opened 2021-04-13
Optimise social feed update

ianopolous:


src/peergos/shared/user/TofuKeyStore.java
```diff
@@ -48,6 +48,13 @@ public TofuKeyStore() {
         return chains.getOrDefault(username, expired.getOrDefault(username, Collections.emptyList()));
     }
 
+    /**
```
kevodwyer, line 51: update the javadocs

src/peergos/shared/user/TofuKeyStore.java
```diff
@@ -57,6 +64,8 @@ public TofuKeyStore() {
                     // we are allowing expired chains to be stored
 
                     List<UserPublicKeyLink> existing = getChain(username);
+                    if (! existing.isEmpty() && existing.get(existing.size() - 1).equals(tail.get(tail.size() - 1)))
```
kevodwyer, line 67: not sure i understand this

ianopolous: This is just saying if nothing has changed for that user in the pki then we don't need to commit the tofukeystore.

src/peergos/shared/user/TofuKeyStore.java
ianopolous: What do you mean? The critical thing here is the return value meaning.

kevodwyer (approved):

PR #820, opened 2021-04-28
Experimental builds in bazel

jandom:
Just building the leaves first (it's easiest to start there – plus it forces a solution to any external deps...)

This is nothing more than a sketch at the moment, the changes proposed for `src/peergos/shared/util/*.java` are very mechanical. 

I would love a hand with java's external deps... I have no idea how to make bazel "eat them up"...


jandom:
Superseded by https://github.com/Peergos/Peergos/pull/828

PR #845, opened 2021-07-02
Implement a private group chat protocol

ianopolous:
This operates without any central servers and without exposing any of the following to the server:
* how many chats a user is in
* the title of the chat
* how many or which members are in a chat
* how many messages are in a chat
* whether a cryptree blob is part of a chat or just a regular file or dir

The model is that every member of a chat maintains an append only log of messages in causal order. To send a new message you just append to your own log. To receive messages you pull from other people's logs and copy into your own mirror. Different logs may have different message orders, but all preserve causal order. Once all messages have been delivered to all recipients, then all members will agree on the state of the chat (it is eventually convergent). 

A chat has private state and shared state. The private state is just the keypair for your identity within the chat, and a list of users who have deleted themselves from the chat without sending a leave message - which is used to stop trying to pull form their log. The shared state is a folder containing the message log, their view of the state of the chat, and any media shared in the chat. 

There is a Map<String, String> of group state which members can set. These mappings form a CRDT with concurrent puts tie-broken by Id. This is used to set the title and the admin list for a chat.

To invite someone to a chat you send an INVITE message anchoring the invite to their pki identity, and then share the shared state with them. To join a chat you copy the shared state into your own space, send a JOIN message, and share this mirror back with the invitee. 

The creator of a chat is an admin. Admins can remove other members, grant admin status to others or remove admin status. Admins also automatically remove members who have deleted their mirror (without cleanly leaving). When someone has left the chat everyone rotates the keys on their mirror and re-shares with remaining members who they are friends with. This means that the members of a chat must form a connected friend graph for message to propagate. We plan to remove this limitation in the future using direct "notification" capabilities that would be shared in the chat and allow sharing chat caps to people you are not friends with (still without any exposure to the server). 

The Ids of each user in a chat form a generalized vector clock, called a TreeClock. 

The corresponding web interface is: https://github.com/Peergos/web-ui/pull/282


src/peergos/server/tests/MultiUserTests.java
```diff
@@ -214,6 +214,21 @@ public void groupSharing() {
...
+    }
+
+    @Test
+    public void chatLeaveAndDelete() {
+        PeergosNetworkUtils.memberLeaveAndDeleteChat(network, random);
+    }
+
+    @Test
+    public void messagingVariations() {
+        PeergosNetworkUtils.editMessage(network, random);
```
kevodwyer, line 229: maybe rename method to editMessage()

src/peergos/server/tests/PeergosNetworkUtils.java
```diff
@@ -1645,6 +1648,215 @@ public static void socialFeedVariations(NetworkAccess network, Random random) {
...
+        FileRef fileRef = ref.get();
+        Optional<FileWrapper> mediaFile = a.getByPath(fileRef.path).join();
+        Assert.assertTrue(mediaFile.isPresent());
+
+        // remove member from chat
+        controllerA = msgA.removeMember(controllerA, b.username).join();
+        controllerA = msgA.sendMessage(controllerA, new ApplicationMessage(Arrays.asList(new Text("B shouldn't see this!")))).join();
+        try {
+            controllerB = msgB.mergeMessages(controllerB, a.username).get();
+            throw new RuntimeException("Fail!");
```
kevodwyer, line 1744: This doesn't seem right. mergeMessages no longer throws Exception

kevodwyer (commented):
well done! incredible amount of work

kevodwyer (approved):

PR #863, opened 2021-09-04
Implement 3rd party identity proofs

ianopolous:
These are just a signed statement saying that you control
this peergos account and this 3rd party account which is posted
to the other service and also stored in Peergos. In Peergos the file 
has a new mimetype application/vnd.peergos-identity-proof

We take it a step further and allow the proof to be encrypted.
This allows you to post a proof to the 3rd party service (which reveals nothing)
and then selectively reveal it by sharing the key. This can be simply sharing
the proof file in Peergos with friends, or out of band. The former is an easy way to prove to other 
Peergos users that you control a given external account. 

An unencrypted proof allows anyone who can see the post on the 3rd party service to proof that the account owner also controls the given Peergos account.


kevodwyer (approved):

src/peergos/server/LinkIdentity.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public static void link(Args a, NetworkAccess network, Crypto crypto) {
+        String username = a.getArg("username");
+        Console console = System.console();
+        String password = new String(console.readPassword("Enter password for " + username + ":"));
+        UserContext context = UserContext.signIn(username, password, network, crypto).join();
+        String usernameB = a.getArg("service-username");
+        String serviceB = a.getArg("service");
+        boolean encrypted = a.getBoolean("encrypted");
+        boolean publish = ! encrypted && a.getBoolean("publish", false);
```
kevodwyer, line 33: Is that logic correct?

src/peergos/server/Main.java
```diff
@@ -402,6 +402,66 @@ public static final void poststrap(Args args) {
...
+                    URL api = new URL(peergosUrl);
+                    NetworkAccess network = Builder.buildJavaNetworkAccess(api, ! peergosUrl.startsWith("http://localhost")).join();
+                    LinkIdentity.link(a, network, crypto);
+                    return true;
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
+                }
+            },
+            Stream.of(
+                      new Command.Arg("peergos-url", "Address of the Peergos server to migrate to", false, "http://localhost:8000"),
```
kevodwyer, line 420: wrong text for arg

src/peergos/server/Main.java
```diff
@@ -402,6 +402,66 @@ public static final void poststrap(Args args) {
...
+                    URL api = new URL(peergosUrl);
+                    NetworkAccess network = Builder.buildJavaNetworkAccess(api, ! peergosUrl.startsWith("http://localhost")).join();
+                    LinkIdentity.verify(a, network);
+                    return true;
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
+                }
+            },
+            Stream.of(
+                      new Command.Arg("peergos-url", "Address of the Peergos server to migrate to", false, "http://localhost:8000"),
```
kevodwyer, line 444: incorrect text

src/peergos/shared/user/IdentityLinkProof.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public String getFilename() {
+        return claim.usernameB + "." + claim.serviceB.name() + ".id.cbor";
+    }
+
+    public String getUrlToPost(FileWrapper proofFile, boolean isPublic) {
+        if (isPublic) {
+            String pathToProof = claim.usernameA + "/.profile/ids/" + getFilename();
+            String path = "/public/" + pathToProof + "?open=true";
+            return "https://beta.peergos.net" + path;
```
kevodwyer, line 70: extract beta.peergos.net out somehow?

src/peergos/server/LinkIdentity.java
ianopolous: An encrypted one shouldn't be published as that would defeat the point of encrypting it.

src/peergos/shared/user/IdentityLinkProof.java
ianopolous: done

kevodwyer (approved):

PR #875, opened 2021-11-07
Switch to IPFS Nucleus

ianopolous:


src/peergos/server/Main.java
```diff
@@ -581,9 +581,9 @@ public static UserService startPeergos(Args a) {
                     a.getBoolean("include-csp", true), basicAuth, webroot, useWebAssetCache, isPublicServer, maxConnectionQueue, handlerThreads);
             boolean isPkiNode = nodeId.equals(pkiServerNodeId);
             if (! isPkiNode && useIPFS) {
-                int pkiNodeSwarmPort = a.getInt("pki.node.swarm.port");
-                InetAddress pkiNodeIpAddress = InetAddress.getByName(a.getArg("pki.node.ipaddress"));
-                ipfsWrapper.connectToNode(new InetSocketAddress(pkiNodeIpAddress, pkiNodeSwarmPort), pkiServerNodeId);
+//                int pkiNodeSwarmPort = a.getInt("pki.node.swarm.port");
```
kevodwyer, line 584: maybe add a note to explain the comment

src/peergos/shared/storage/ContentAddressedStorage.java
```diff
@@ -593,21 +593,17 @@ private static String encode(String component) {
 
         @Override
         public CompletableFuture<List<Multihash>> recursivePin(PublicKeyHash owner, Multihash hash) {
-            return poster.get(apiPrefix + PIN_ADD + "?stream-channels=true&arg=" + hash.toString()
-                    + "&owner=" + encode(owner.toString())).thenApply(this::getPins);
+            return Futures.of(Arrays.asList(hash));
```
kevodwyer, line 596: can you explain these changes?

ianopolous: We're not calling pin on ipfs at all anymore. This was only relevant when we were using the built-in ipfs GC, but we haven't been doing that for a while now. Pinning stops data from being garbage collected by the ipfs GC.

src/peergos/server/Main.java
ianopolous: done

kevodwyer (approved):

PR #882, opened 2021-12-15
Add JsMethod for calculateThumbnail

kevodwyer:


src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -1000,6 +1000,27 @@ public static int getNumberOfChunks(long size) {
                 );
     }
 
+    @JsMethod
+    public CompletableFuture<Boolean> calculateThumbnail(NetworkAccess network, Crypto crypto) {
```
ianopolous, line 1004: calculate suggests to me that something is returned. Maybe "updateThumbnail"?

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -1000,6 +1000,27 @@ public static int getNumberOfChunks(long size) {
...
+
+        BiFunction<Snapshot, Committer, CompletableFuture<Pair<Snapshot, Boolean>>> updateThumbnail =
```
ianopolous, line 1006: Any reason not to inline this lambda?

kevodwyer: now inlined

src/peergos/shared/user/fs/FileWrapper.java
kevodwyer: updateThumbnail is already taken (used as part of updateProfile)

ianopolous: That's annoying, maybe calculateAndUpdateThumbnail? Probably existing updateThumbnail should have been called setThumbnail

kevodwyer: done

ianopolous (approved):

Issue #862, opened 2021-09-02
Implement BAT based block access control


ianopolous:
This is an additional layer of protection at the lowest level. This means you need a secret token to retrieve a block at all. 

Assume an auth token derived from a secret block access token (BAT) is sent along with each request for a block (either via web-ui or bitswap). 

Needs to be backwards compatible with existing blocks until users have logged in and upgraded their blocks, which can be done gradually. It also needs to preserve O(1) seeking in large files.

The auth scheme will be the same as S3 (V4) where the signed URL includes {requested cid, requesting nodeId, expiry, BAT ID}. Note the p2p ipfs connections are TLS 1.3 authenticated with the nodeId key pair, so the requestor can't be faked) Bitswap's request will send such an auth string, and thus the ipfs api will need to accept it. The peergos api can just take the BAT token itself (it will see it anyway once it retrieves the block). Using the BAT itself in the Peergos API also prevents replay attacks on a node using auth string it has sent out on the DHT.

The primary BAT for a block is derivable from the block itself so the server can do stateless auth with just the individual block's data. For the purposes of the signature the IPFS API URL would be:
URL: PUT sourcePeerID/api/v0/block/get?arg=$CID&auth=$AUTH
where $AUTH=$BATID:$EXPIRY:$SIG, which serializes to 89 bytes. The BATID is the sha256(BAT) as a mulithash.

The public API would require passing in the actual BAT (the server will see it anyway once it has the block, by design)
URL: https://public-domain/api/v0/block/get?arg=$CID&bat=$BAT

## Format and derivation
1. All BATs are 32 bytes
2. For raw blocks the BAT will be HKDF(block, "peergos", 32)
3. For cbor nodes the BAT will be random
4. cbor nodes will have 2 BATs, one for fine grained control, and one mirror one which is per user

## BAT storage
1. for raw blocks the BATs are only stored in the referencing cryptree node (but the server doesn't need the cryptree node to verify incoming requests, just the block itself)
2. for cbor nodes all BATs are stored in a list in the root object under key "bats"
3. The primary (fine-grained) BAT for a block is stored in the clear in this list.
4. Any more privileged BATs (e.g. the user wide "mirror" BAT) the hash of the BAT is stored in the block, and the server maintains a mapping from hash => privileged BAT for each user. 

## Cryptree BAT structure
| | Chunk 1                |  Chunk 2     | Chunk 3 |
|-|-------------------------|-----------------|---------|
[...]

PR #884, opened 2021-12-31
Implement post-quantum block level access control

ianopolous:
This adds a capability based scheme for authorising retrieval of a block. To retrieve a block you need a Block Access Token (BAT) which is 32 random bytes. To retrieve a block you use an applicable BAT to generate an S3 V4 signature and include that in the request over bitswap (or send the BAT directly to the Peergos API). The signature incorporates the nodeId of the sender in place of the domain name for the target for S3. This allows us to broadcast the same signature to the DHT and each node can verify it was sent from the actual signing node, and protects against replay attacks. 

Cryptree nodes store a list of BATs under the cbor key "bats" in the top level. They also store a nested BAT for each linked raw fragment. Each user has a mirror BAT which can be used to mirror their entire block set. A server can also have a server wide BAT which can be used to retrieve any block on the instance. More details on BATs are here: https://github.com/Peergos/Peergos/issues/862

For files, subsequent chunk BATs are derived using the stream secret in the same way as the map keys. This allows us to keep the O(1) seeking in huge files (just local hashing). 

Closes: https://github.com/Peergos/Peergos/issues/862


src/peergos/server/storage/S3BlockStorage.java
```diff
@@ -94,14 +100,23 @@ private Multihash keyToHash(String key) {
...
-    public CompletableFuture<List<PresignedUrl>> authReads(List<Multihash> blocks) {
+    public CompletableFuture<List<PresignedUrl>> authReads(List<MirrorCap> blocks) {
         if (blocks.size() > 50)
             throw new IllegalStateException("Too many reads to auth!");
         List<PresignedUrl> res = new ArrayList<>();
 
-        for (Multihash block : blocks) {
-            String s3Key = hashToKey(block);
-            res.add(S3Request.preSignGet(s3Key, Optional.of(600), ZonedDateTime.now(), host, region, accessKeyId, secretKey));
+        // retrieve all blocks and verify BATs in aprallel
```
kevodwyer, line 108: spelling of parallel

src/peergos/shared/storage/auth/Bat.java
```diff
@@ -0,0 +1,100 @@
...
+            throw new IllegalStateException("Cannot use identity multihash in S3 signatures!");
+        S3Request req = new S3Request("GET", sourceNode.toBase58(), "api/v0/block/get?arg=" + block.toBase58(), S3Request.UNSIGNED,
+                Optional.of(expirySeconds), false, true,
+                Collections.emptyMap(), Collections.emptyMap(), batId.toBase58(), "eu-central-1", datetime);
+        return S3Request.computeSignature(req, encodeSecret(), h)
+                .thenApply(signature -> new BlockAuth(ArrayOps.hexToBytes(signature), expirySeconds, datetime, batId));
+    }
+
+    public static Bat deriveFromRawBlock(byte[] block) {
+        // TODO need to analyse the cryptographic properties of this construction, might need to use hmac_sha256 instead
```
kevodwyer, line 85: this still a todo?

kevodwyer (changes requested):
I didn't mark all the places with the word Cave. I would suggest dropping it.

src/peergos/shared/storage/auth/Bat.java
ianopolous: yep.

src/peergos/server/Main.java
```diff
@@ -770,7 +803,9 @@ public static boolean migrate(Args a) {
             }
             System.out.println("Migrating user from node " + currentStorageNodeId + " to " + newStorageNodeId);
             List<UserPublicKeyLink> newChain = Migrate.buildMigrationChain(existing, newStorageNodeId, user.signer.secret);
-            user.network.coreNode.migrateUser(username, newChain, currentStorageNodeId).join();
+            List<BatWithId> mirrorBats = user.network.batCave.getUserBats(username, user.signer).join();
+            BatWithId current = mirrorBats.get(mirrorBats.size() - 1);
```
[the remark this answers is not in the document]
ianopolous: done

kevodwyer (approved):

PR #893, opened 2022-01-30
block p2p http get calls

ianopolous:
We do this by listening on an additional port which the p2p http requests are proxied to. This can then ignore block.get calls.


kevodwyer (approved):
Looks good

PR #894, opened 2022-01-31
Updated - Do not recalculate thumbnail as part of copy/move

kevodwyer:


src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -1450,15 +1457,16 @@ public SigningPrivateKeyAndPublicHash signingPair() {
...
                 .thenApply(newAccess -> true);
     }
 
     public CompletableFuture<Snapshot> copyTo(FileWrapper target,
                                               NetworkAccess network,
                                               Crypto crypto,
                                               Snapshot version,
-                                              Committer committer) {
+                                              Committer committer,
+                                              Optional<Thumbnail> existingThumbnail) {
```
ianopolous, line 1469: The convention I have at the moment is to keep "meaningful" parameters before things like network, crypto, version and committer. Could we put exisitngThumnail as second param?

src/peergos/shared/user/fs/FileWrapper.java
```diff
@@ -1091,16 +1092,22 @@ public static int getNumberOfChunks(long size) {
...
+        return network.getFile(base, cap, getChildsEntryWriter(), ownername).thenCompose(fileOpt -> {
+            if (fileOpt.get().props.thumbnail.isEmpty()) {
+                return generateThumbnail(network, fileData, (int) Math.min(fileSize, Integer.MAX_VALUE), fileName, mimeType)
+                        .thenCompose(thumbData -> {
+                            if (thumbData.isEmpty())
+                                return Futures.of(base);
+                            FileProperties fileProps = new FileProperties(fileName, false, props.isLink, mimeType, fileSize,
+                                    updatedDateTime, isHidden, thumbData, streamSecret);
+
+                            return network.getFile(base, cap, getChildsEntryWriter(), ownername)
```
ianopolous, line 1104: no need to get the file again here

ianopolous (approved):

PR #902, opened 2022-02-21
Optimise delete

ianopolous:
Use a buffered network which makes delete 7.5-10X faster

### DIFF
--- a/src/peergos/shared/user/SharedWithCache.java
+++ b/src/peergos/shared/user/SharedWithCache.java
@@ -273,7 +273,7 @@ public class SharedWithCache {
                 .thenApply(opt -> opt.map(s -> s.get(getFilename(p))).orElse(FileSharedWithState.EMPTY));
     }
 
-    public CompletableFuture<Snapshot> applyAndCommit(Path toFile, Function<SharedWithState, SharedWithState> transform, Snapshot in, Committer committer) {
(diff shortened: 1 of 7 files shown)


src/peergos/shared/user/SharedWithCache.java
```diff
@@ -273,19 +290,39 @@ private static String getFilename(Path p) {
...
                     SharedWithState updated = transform.apply(current);
+                    if (current.equals(updated))
+                        return Futures.of(p.left.version);
                     byte[] raw = updated.serialize();
                     return source.overwriteFile(AsyncReader.build(raw), raw.length, network, crypto, x -> {}, source.version, committer);
                 });
     }
 
-    public CompletableFuture<Snapshot> rename(Path initial, Path after, Snapshot in, Committer committer) {
+    public CompletableFuture<Snapshot> applyIfPresentAndCommit(Path toFile, Function<SharedWithState, SharedWithState> transform, Snapshot in, Committer committer, NetworkAccess network) {
```
kevodwyer, line 307: Having network as a param in a public method and as a class member is confusing (especially since it is used everywhere). Sure, it is necessary to actually use the BufferedNetworkAccess for remove. I don't have a good alternate suggestion.

ianopolous: Yep. I'm starting to realise that storing NetworkAccess everywhere is an anti pattern. I can try and remove the instance var in another PR, after the one after this.

kevodwyer (approved):

Issue #924, opened 2022-04-11
Unauthorized! Error


DillFrescott:
java.util.concurrent.CompletionException: java.lang.IllegalStateException: Unauthorised!
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315)
        at java.base/java.util.concurrent.CompletableFuture.uniApplyNow(CompletableFuture.java:687)
        at java.base/java.util.concurrent.CompletableFuture.uniApplyStage(CompletableFuture.java:662)
        at java.base/java.util.concurrent.CompletableFuture.thenApply(CompletableFuture.java:2168)
        at peergos.server.storage.AuthedStorage.getRaw(AuthedStorage.java:58)
        at peergos.server.storage.AuthedStorage.getRaw(AuthedStorage.java:54)
        at peergos.server.storage.AuthedStorage.get(AuthedStorage.java:44)
        at peergos.server.storage.DeletableContentAddressedStorage.lambda$get$0(DeletableContentAddressedStorage.java:50)
        at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1187)
        at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2309)
        at peergos.server.storage.DeletableContentAddressedStorage.get(DeletableContentAddressedStorage.java:50)
        at peergos.server.storage.AuthedStorage.get(AuthedStorage.java:39)
        at peergos.server.storage.AuthedCachingStorage.get(AuthedCachingStorage.java:107)
        at peergos.shared.storage.DelegatingStorage.get(DelegatingStorage.java:54)
        at peergos.shared.storage.ContentAddressedStorage$Proxying.get(ContentAddressedStorage.java:578)
        at peergos.server.storage.AuthedCachingStorage.get(AuthedCachingStorage.java:107)
        at peergos.server.net.DHTHandler.handle(DHTHandler.java:211)
        at peergos.server.net.SubdomainHandler.handle(SubdomainHandler.java:28)
        at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:95)
        at jdk.httpserver/sun.net.httpserver.AuthFilter.doFilter(AuthFilter.java:82)
        at jdk.httpserver/com.sun.net.httpserver.Filter$Chain.doFilter(Filter.java:98)
[...]

PR #935, opened 2022-04-19
Fix timezone issue which was causing bat auth to fail

ianopolous:
This meant self-hosted sign ins far away from UTC were failing with Unauthorised

Also log auth errors with a reason.

Fixes https://github.com/Peergos/Peergos/issues/924

### DIFF
--- a/src/peergos/server/storage/auth/BlockRequestAuthoriser.java
+++ b/src/peergos/server/storage/auth/BlockRequestAuthoriser.java
@@ -18,14 +18,7 @@ public interface BlockRequestAuthoriser {
         S3Request req = new S3Request("GET", sourceNode.toBase58(), "api/v0/block/get?arg=" + block.toBase58(), S3Request.UNSIGNED,
                 Optional.of(auth.expirySeconds), false, true,
                 Collections.emptyMap(), Collections.emptyMap(), auth.batId.toBase58(), "eu-central-1", t);
-        LocalDateTime timestamp = LocalDateTime.of(
-                Integer.parseInt(t.substring(0, 4)),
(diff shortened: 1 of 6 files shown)


src/peergos/server/storage/auth/BlockRequestAuthoriser.java
```diff
@@ -18,12 +18,38 @@ static boolean isValidAuth(BlockAuth auth, Cid block, Cid sourceNode, Bat bat, H
         S3Request req = new S3Request("GET", sourceNode.toBase58(), "api/v0/block/get?arg=" + block.toBase58(), S3Request.UNSIGNED,
                 Optional.of(auth.expirySeconds), false, true,
                 Collections.emptyMap(), Collections.emptyMap(), auth.batId.toBase58(), "eu-central-1", t);
-        Instant timestamp = Instant.parse(String.format("%s-%s-%sT%s:%s:%sZ", t.substring(0, 4), t.substring(4, 6), t.substring(6, 8), t.substring(9, 11), t.substring(11, 13), t.substring(13, 15)));
-        Instant expiry = timestamp.plusSeconds(auth.expirySeconds);
-        if (expiry.isBefore(Instant.now()))
+        LocalDateTime timestamp = LocalDateTime.of(
```
kevodwyer, line 21: have you tested this with leading zeros?

ianopolous: It's parsing as an integer, so leading zeroes don't matter.

ianopolous: done

kevodwyer (approved):

Issue #932, opened 2022-04-18
How to use fuse mount on Windows


jonaswre:
I've downloaded the binary. And ./peergos.exe fuse -username "myusername" -password "mypassword"

But I always get "Connection refused: connect retrieving peergos/v0/core/getUsernamesGzip/" 

Do I need to specify webport and mountPoint?

ianopolous:
Yes the fuse command assumes a Peergos daemon is already running locally. The FUSE integration hasn't been tested on Windows so do let us know how you get on.

jonaswre:
Thanks for the info. 

I now get the error 

> Apr 18, 2022 5:45:13 PM jnr.ffi.util.ref.internal.Finalizer getInheritableThreadLocalsField INFO: Couldn't access Thread.inheritableThreadLocals. Reference finalizer threads will inherit thread local values.

 in the fuse mount bash and 

> java.util.concurrent.CompletionException: java.net.SocketTimeoutException: Read timed out
        at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:332)
        at java.base/java.util.concurrent.CompletableFuture.uniApplyNow(CompletableFuture.java:674)
        at java.base/java.util.concurrent.CompletableFuture.uniApplyStage(CompletableFuture.java:662)
        at java.base/java.util.concurrent.CompletableFuture.thenApply(CompletableFuture.java:2168)
        at peergos.shared.storage.ContentAddressedStorageProxy$HTTP.getChampLookup(ContentAddressedStorageProxy.java:100)
        at peergos.shared.storage.ContentAddressedStorage$Proxying.lambda$getChampLookup$5(ContentAddressedStorage.java:573)
        at peergos.shared.corenode.Proxy.lambda$redirectCall$0(Proxy.java:28)
        at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1187)
        at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2309)
        at peergos.shared.corenode.Proxy.lambda$redirectCall$1(Proxy.java:19)
        at java.base/java.util.concurrent.CompletableFuture.uniComposeStage(CompletableFuture.java:1187)
        at java.base/java.util.concurrent.CompletableFuture.thenCompose(CompletableFuture.java:2309)
        at peergos.shared.corenode.Proxy.redirectCall(Proxy.java:18)
[...]

PR #945, opened 2022-04-30
Update README.md

jonaswre:
In refrence to #932 

This might be useful for other people. Especially the http/https part wasn't obvious to  me.


ianopolous (approved):
Awesome! Thank you!

PR #946, opened 2022-05-02
Peergos Apps

kevodwyer:
web-ui PR: https://github.com/Peergos/web-ui/pull/384


src/peergos/server/net/StaticHandler.java
```diff
@@ -115,7 +115,7 @@ else if (path.endsWith(".svg"))
                         " " + this.host +
                         (isSubdomain ? " 'unsafe-inline' https://" + reqHost : "") + // calendar, editor, todoboard, pdfviewer
                         ";" +
-                        (isSubdomain ? "sandbox allow-scripts allow-forms;" : "") +
+                        (isSubdomain ? "sandbox allow-same-origin allow-scripts allow-forms;" : "") +
```
ianopolous, line 118: We can't have allow-same-origin here as that would allow direct calls to the api.

PR #969, opened 2022-07-22
app dev proxy

kevodwyer:
UI change: https://github.com/Peergos/web-ui/pull/404


jonaswre:
@ianopolous can you provide me a latest jar?
I think i'll need to setup my own build pipeline but im corrently not on my main computer. 
Or maybe you can add artifacts to the build action that way I can always just grab the latest from the build.

PR #973, opened 2022-09-08
Fix usage growth

ianopolous:
This fixes several issues creating space usage growth unnecessarily. 

There is a UserGC util to clear all unreachable champ nodes (unreachable from the filesystem). 

Champ removals now fully remove the mapping, not just set the value to empty. This need mutable pointers to have a sequence number to avoid replay attacks. 

File upload transaction files were not being deleted, just unlinked from parent dir. 

Now there is a test to enforce upload + mkdir followed by delete doesn't increase space usage.


ianopolous:
Longs are fine internally (as long as they don't cross the JSNI boundary) - "As of GWT 1.5, long primitives are emulated as a pair of 32-bit integers, and work reliably over the entire 64-bit range."

PR #981, opened 2022-10-11
Offline client

ianopolous:
Add a bunch of util to enable an offline client. 

For now these are used in Android build, but could also be used in browser without sqlite.

### DIFF
--- a/src/peergos/server/OfflineBatCache.java
+++ b/src/peergos/server/OfflineBatCache.java
@@ -1,0 +1,39 @@
+package peergos.server;
+
+import peergos.server.storage.auth.*;
+import peergos.shared.storage.auth.*;
+import peergos.shared.util.*;
+
+import java.util.*;
+import java.util.concurrent.*;
+
+public class OfflineBatCache implements BatCave {
+
+    private final BatCave target;
+    private final BatCache cache;
+
+    public OfflineBatCache(BatCave target, BatCache cache) {
+        this.target = target;
+        this.cache = cache;
+    }
+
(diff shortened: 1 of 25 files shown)


src/peergos/server/OfflineBatCache.java
```diff
@@ -0,0 +1,39 @@
...
+                () -> target.getUserBats(username, auth).thenApply(bats -> {
+                    cache.setUserBats(username, bats);
+                    return bats;
+                }),
+                t -> cache.getUserBats(username));
+    }
+
+    @Override
+    public CompletableFuture<Boolean> addBat(String username, BatId id, Bat bat, byte[] auth) {
+        return target.addBat(username, id, bat, auth);
```
kevodwyer, line 37: add to cache if successful

src/peergos/server/util/JavaPoster.java
```diff
@@ -52,6 +52,7 @@ private CompletableFuture<byte[]> post(String url, byte[] payload, boolean unzip
             conn = (HttpURLConnection) buildURL(url).openConnection();
             if (timeoutMillis >= 0)
                 conn.setReadTimeout(timeoutMillis);
+            conn.setConnectTimeout(1_000);
```
kevodwyer, line 55: seems low. debugging code?

ianopolous: 1s is huge for connection latency. UK to Aus should be < .3s.

kevodwyer (approved):

PR #1001, opened 2022-11-23
Detect html mimetype based on content as well

ianopolous:


kevodwyer (commented):
add trim() and toLowerCase() before contains test

kevodwyer (approved):

PR #1020, opened 2022-12-23
Fix mirroring a user's data on an S3 instance

ianopolous:
Errors were not being logged.

Give DeletableCAS an explicit option to not check auth in internal calls as all implementations had this anyway. Use this to not enforce the auth during mirroring as we might not have the mirror bat committed locally.

Bump version.

### DIFF
--- a/src/peergos/server/Builder.java
+++ b/src/peergos/server/Builder.java
@@ -279,7 +279,8 @@ public class Builder {
                                 return ALLOW;
                         }
                         if (! batids.isEmpty()) {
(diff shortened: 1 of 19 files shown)


kevodwyer (approved):

kevodwyer (commented):
getOrMkdirs + mkdir is used in web-ui code. Have you confirmed they are also correct?

ianopolous:
As long as the UI passes in the user's mirror bat then they are correct.

kevodwyer (approved):

PR #1039, opened 2023-04-11
Allow listen address to be configured.

ianopolous:
If we are not listening on localhost only then disable open signups, and require token usage to signup.


src/peergos/server/Main.java
```diff
@@ -762,14 +764,14 @@ public static UserService startPeergos(Args args) {
...
-            boolean generateToken = a.getBoolean("generate-token", false);
+            boolean generateToken = a.getBoolean("generate-token", ! localhostApi);
             if (generateToken) {
                 System.out.println("Generating signup token...");
                 String token = userQuotas.generateToken(crypto.random);
                 System.out.println("Peergos daemon started. Browse to http://localhost:" + webPort + "/?signup=true&token="
                         + token + " to sign up.");
             } else
-                System.out.println("Peergos daemon started. Browse to http://localhost:" + webPort + "/ to sign up or login. ");
+                System.out.println("Peergos daemon started. Browse to http://localhost:" + webPort + "/ to sign up or login. \nRun with -generate-token true to generate a signup token.");
```
gsaslis, line 774: can we remove the hard-coded `localhost` as part of this change ?

ianopolous: Done

PR #1040, opened 2023-04-12
Support wildcard listener address

ianopolous:
This is mainly useful in docker where proxied connections come in on random non local ips

So you can run with -domain 0.0.0.0 -public-domain localhost:8000
if docker is set to proxy localhost:8000 to peergos.


kevodwyer (approved):
Would be good to document this in README.md

PR #1070, opened 2023-08-04
Use direct calls to Nabu rather then through JavaPoster

kevodwyer:


src/peergos/server/storage/NabuStorage.java
```diff
@@ -0,0 +1,103 @@
...
+                                               List<byte[]> blocks,
+                                               String format,
+                                               TransactionId tid) {
+        for (byte[] block : blocks) {
+            if (block.length > MAX_BLOCK_SIZE)
+                throw new IllegalStateException("Invalid block size: " + block.length
+                        + ", blocks must be smaller than 1MiB!");
+        }
+        List<Cid> cidList = new ArrayList<>(blocks.size());
+        for (byte[] block : blocks) {
```
ianopolous, line 50: This should use a bulk call, if we don't have one, we need one.

src/peergos/server/storage/NabuStorage.java
```diff
@@ -0,0 +1,103 @@
...
+import peergos.shared.storage.auth.BatWithId;
+import peergos.shared.user.HttpPoster;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+
+public class NabuStorage extends ContentAddressedStorage.HTTP {
```
ianopolous, line 23: Why does this extend HTTP?

ianopolous: But a nabi api is by definition never a peergos server

src/peergos/server/storage/NabuStorage.java
ianopolous: Is block.put not a multipart request?

ianopolous: I see it is. We should add a bulk put to Blockstore interface

PR #1085, opened 2023-09-21
fix for CAS error when posting on newsfeed

kevodwyer:


src/peergos/shared/social/SocialFeed.java
```diff
@@ -220,12 +220,13 @@ private CompletableFuture<Boolean> mergeCommentsIntoParent(String parentPath,
 
     private synchronized CompletableFuture<Boolean> commit() {
         byte[] raw = new FeedState(lastSeenIndex, feedSizeRecords, feedSizeBytes, currentCapBytesProcessed).serialize();
-        return stateFile.overwriteFile(AsyncReader.build(raw), raw.length,
-                network, crypto, x -> {})
-                .thenApply(f -> {
-                    this.stateFile = f;
-                    return true;
-                });
+        return this.stateFile.getLatest(network).thenCompose(updatedStateFile ->
```
ianopolous, line 223: Can we exercise this in one of the unit tests?

ianopolous: We should be getting an updated SocialFeed after adding the post, so this shouldn't be necessary here.

PR #1097, opened 2023-10-24
inline BlockRequestAuthoriser calls

kevodwyer:
This allows IPFS to evaluate auth on block requests directly, and so we can remove the block auth server in peergos (one less listening port). 
IPFS only runs will need the bat db config added,


src/peergos/server/Main.java
```diff
@@ -540,6 +540,9 @@ public static UserService startPeergos(Args a) {
 
             BatCave batStore = new JdbcBatCave(getDBConnector(a, "bat-store", dbConnectionPool), sqlCommands);
             BlockRequestAuthoriser blockRequestAuthoriser = Builder.blockAuthoriser(a, batStore, hasher);
+            if (useIPFS) {
+                ipfsWrapper.setBlockRequestAuthoriser(blockRequestAuthoriser);
```
ianopolous, line 544: what happens to reqs it receives before this is set?

src/peergos/server/storage/IpfsWrapper.java
```diff
@@ -234,6 +242,9 @@ public BlockMetadataStore getBlockMetadata() {
         return metaDB;
     }
 
+    public void setBlockRequestAuthoriser(BlockRequestAuthoriser blockRequestAuthoriser) {
```
ianopolous, line 245: Is it possible to avoid this method and set it at construction time?

src/peergos/server/Main.java
ianopolous: It's still a race condition as ipfs is started and listening by then. What happens to those requests?

ianopolous:
Does this also set the block auth function when the CLI runs the IPFS command?

src/peergos/server/tests/linux/S3UserTests.java
```diff
@@ -29,6 +29,7 @@ public class S3UserTests extends UserTests {
 
     private static Args pkiArgs = buildArgs()
             .with("useIPFS", "true")
+            .with("transactions-sql-file", "transactions.sql")
```
ianopolous, line 32: This should have a default value no? Then we don't need to set it to the same value everywhere?

src/peergos/server/tests/UserTests.java
```diff
@@ -88,7 +88,9 @@ public static Args buildArgs() {
                     Main.PEERGOS_PATH, peergosDir.toString(),
                     "peergos.password", "testpassword",
                     "pki.keygen.password", "testpkipassword",
-                    "pki.keyfile.password", "testpassword"
+                    "pki.keyfile.password", "testpassword",
+                    "transactions-sql-file", "transactions.sql",
```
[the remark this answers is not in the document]
ianopolous: Scratch that. This is a new parameter. So kinda dangerous to have a default which would likely not point to the correct thing.

src/peergos/server/storage/IpfsWrapper.java
```diff
@@ -207,7 +208,16 @@ public static IpfsConfigParams buildConfig(Args args) {
 
     private final BlockMetadataStore metaDB;
 
-    public IpfsWrapper(Path ipfsDir, IpfsConfigParams ipfsConfigParams, BlockMetadataStore metaDB) {
+    private final BlockRequestAuthoriser blockRequestAuthoriser;
+
+    private final Supplier<Connection> dbConnectionPool;
```
ianopolous, line 213: Aren't the batcave and connpool implementation details of the authoriser? Why do they need to be member fields?

ianopolous: I see why they're there now, but it feels weird and tightly coupled.

ianopolous (approved):

PR #1104, opened 2023-11-02
save IPFS identity when creating a new identity

kevodwyer:


src/peergos/server/storage/IpfsWrapper.java
```diff
@@ -235,12 +237,24 @@ public IpfsWrapper(Path ipfsDir, IpfsConfigParams ipfsConfigParams) {
...
+                identityOpt = Optional.of(newIdentity);
             }
         }
         this.ipfsConfigParams= ipfsConfigParams.withIdentity(identityOpt);
     }
 
+    private void saveIdentityToFile(IdentitySection identity) {
+        String text = JsonHelper.pretty(identity.toJson());
+        try {
+            Path ipfsDir = this.ipfsDir.resolve("identity.json");
```
ianopolous, line 251: Can we do this in the same format as existing ipfs nodes for consistency?

PR #1105, opened 2023-11-02
add identity to peergos config if read from ipfs config or generated

kevodwyer:


src/peergos/server/Main.java
```diff
@@ -708,7 +708,14 @@ public static UserService startPeergos(Args a) {
                         .with("peergos-url", "http://localhost:" + a.getArg("port"));
                 GATEWAY.main(gatewayArgs);
             }
-            a.saveToFileIfAbsent();
+
+            if (useIPFS && !a.hasArg("ipfs.identity.peerid")) {
+                Args args = a.with("ipfs.identity.peerid", ipfsWrapper.ipfsConfigParams.identity.get().peerId.toBase58());
+                args = args.with("ipfs.identity.priv-key", Base64.getEncoder().encodeToString(ipfsWrapper.ipfsConfigParams.identity.get().privKeyProtobuf));
+                args.saveToFile();
```
ianopolous, line 715: This is going to overwrite the config every run. Can we leave the writing as it was?

ianopolous (approved):

Issue #1181, opened 2024-09-14
[bug] Offline login to a local proxy does not work


rapenne-s:
My setup is the following, I have a Peergos instance on my LAN that is available over the Internet, and I made a local proxy on my laptop (using http://127.0.0.1:8000 with `java -jar Peergos.jar daemon`).

When I disconnect the laptop from the network, I am not able to login on http://127.0.0.1:8000 using my credentials. After some time (30 seconds maybe), I have a red popup with "Network timeout"

rapenne-s:
The local proxy logs when trying to connect when offline

```
$ java -jar Peergos.jar daemon
Couldn't load native crypto library at /var/home/solene/Downloads/native-lib/libtweetnacl.so, using pure Java version...
To use the native linux-x86-64 crypto implementation use option -Djava.library.path=native-lib
Starting Peergos daemon version: 0.19.0-web-ui/3ef5e383e9fe01ee2710b13ab15b7e733bae5e52
sept. 14, 2024 3:32:59 PM io.netty.util.internal.MacAddressUtil defaultMachineId
AVERTISSEMENT: Failed to find a usable hardware address from the network interfaces; using random bytes: 35:e3:d0:7f:64:36:2b:b8

█╗█╗█╗█╗   ██████╗ ███████╗███████╗██████╗  ██████╗  ██████╗ ███████╗   █╗█╗█╗█╗
 █████╔╝   ██╔══██╗██╔════╝██╔════╝██╔══██╗██╔════╝ ██╔═══██╗██╔════╝    █████╔╝
 ██ ██║    ██████╔╝█████╗  █████╗  ██████╔╝██║  ███╗██║   ██║███████╗    ██ ██║
 █████║    ██╔═══╝ ██╔══╝  ██╔══╝  ██╔══██╗██║   ██║██║   ██║╚════██║    █████║
███████╗   ██║     ███████╗███████╗██║  ██║╚██████╔╝╚██████╔╝███████║   ███████╗
╚══════╝   ╚═╝     ╚══════╝╚══════╝╚═╝  ╚═╝ ╚═════╝  ╚═════╝ ╚══════╝   ╚══════╝
Peergos daemon started. Browse to http://localhost:8000/ to sign up or login. 
Run with -generate-token true to generate a signup token.
Running version 0.19.0-web-ui/3ef5e383e9fe01ee2710b13ab15b7e733bae5e52
[...]

PR #1191, opened 2024-09-16
Optimise offline login when connected to a daemon

ianopolous:
(e.g. localhost when host has no network)
We race the cache for pointer, login and bat gets. Update the cache on successful remote calls.

Fixes https://github.com/Peergos/Peergos/issues/1181


ianopolous:
CompletableFuture values cannot change once they are set. Java throws in this case. It wouldn't matter anyway because we only ever have one 'thenable' dependency. It works fine in the browser - login is normal speed offline on a local proxy.

PR #1195, opened 2024-09-18
increase timeout back to 30 seconds

kevodwyer:
web ui PR: https://github.com/Peergos/web-ui/pull/617


src/peergos/shared/user/HttpPoster.java
```diff
@@ -8,13 +8,13 @@ public interface HttpPoster {
     CompletableFuture<byte[]> post(String url, byte[] payload, boolean unzip, int timeoutMillis);
 
     default CompletableFuture<byte[]> post(String url, byte[] payload, boolean unzip) {
-        return post(url, payload, unzip, 15_000);
+        return post(url, payload, unzip, 30_000);
```
ianopolous, line 11: Can we do this in a way that only affect JS and not the server? Or is that already the case?

ianopolous (approved):

PR #1223, opened 2024-12-05
Improve s3 block buffer

ianopolous:
* make it back off and retry


ianopolous:
It's called getWithBackOff because it takes a Supplier and calls .get on it.

PR #1265, opened 2025-03-21
Block external logins by default, unless we are on localhost

ianopolous:


rapenne-s:
I see it's !public-server

What is a public server exactly?

ianopolous:
! public-server is the default value for this new arg. This means you expect the api to be available publicly. So not something running on localhost without a reverse proxy.

Issue #1272, opened 2025-04-11
Control proxy via web-ui


ianopolous:
Add endpoints and ui for changing the proxy target on localhost (before login). 

And for getting/setting the cache size (in the current place).

These should work on android as well.

PR #1358, opened 2026-04-13
feat: support self-hosted servers in desktop/proxy mode

hmarquesgit:
## Summary

The desktop app and `proxy` command currently default to `peergos.net` with no way for the user to persistently choose a different server on first launch. Self-hosted operators have to pass `-peergos-url` on every invocation, and the packaged native app exposes no such flag at all.

This PR adds:

1. **`server-url` arg + config file support** — if `~/.peergos/config` contains `server-url`, the proxy uses it instead of the hardcoded default. CLI flags still take precedence.
2. **`server-url/get` and `server-url/set` HTTP API** in `ConfigHandler` — the web UI can read and persist the server URL without touching the filesystem directly. Localhost-only, with URL validation (https required for non-loopback hosts).
3. **`LocalAppProperties`** in `UserService` — carries the peergos dir path and current server URL to `ConfigHandler`.

### What changed

| File | Change |
|------|--------|
| `Main.java` | New `ARG_SERVER_URL`; `getAppServerUrl()` reads CLI → config → default; wired into PROXY and MAIN commands |
| `UserService.java` | `LocalAppProperties` record; constructor overload to pass it to `ConfigHandler` |
| `ConfigHandler.java` | `server-url/get` and `server-url/set` endpoints; config read/write; URL validation |
| `README.md` | Usage instructions for `-server-url` and config file |

### Backwards compatibility

- **No breaking changes.** Default remains `https://peergos.net`.
- The `peergos-url` arg still works; `server-url` takes precedence when both are present.
- Config file format (`key = value`) is unchanged.

### Companion PR

Frontend changes (server URL panel on login screen, settings option, loopback mixin): https://github.com/Peergos/web-ui — same branch name `feat/self-hosted-desktop-app`.

### How to test

```bash
# CLI flag
java -jar Peergos.jar -server-url https://my.server.com

# Config file
echo "server-url = https://my.server.com" >> ~/.peergos/config
java -jar Peergos.jar

# The login screen on localhost shows a "Desktop server" panel to configure this via UI.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,16 @@ You can then access the web interface and login via http://localhost:8000.
 
 In this mode of operation all your writes are proxied directly to your home server. The local instance caches any blocks you access for faster subsequent access. 
 
+If you are using the packaged desktop app, or the default `peergos` launcher on macOS, you can point it at a self-hosted instance with:
+```bash
+peergos -server-url https://YOUR_PEERGOS_SERVER_DOMAIN
+```
+To make that persistent for the desktop app, add the following to `~/.peergos/config`:
+```ini
+server-url = https://YOUR_PEERGOS_SERVER_DOMAIN
+```
+For security, prefer `https` for remote servers, and only use plain `http` for loopback addresses such as `http://localhost:8000`.
+
 Usage - self hosting
 -----
 Use this method to run a new home-server (which is best with a publicly routable IP, and always on machine) to create accounts on or migrate accounts to.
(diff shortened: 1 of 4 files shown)


ianopolous:
Thank you. The peergos-url arg should already work from the config for persisting this, including the desktop apps.

ianopolous:
https://github.com/Peergos/Peergos/issues/1272

PR #1363, opened 2026-04-21
Close FileOutputStream in get

tejas-ae:
The current code in src/peergos/server/cli/CLI.java likely works fine most of the time, but The resource opened there can leak if get exits on an error path. this patch moves the allocation into try-with-resources so cleanup happens on every exit path. Close this if it’s not worth it.
Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


ianopolous (approved):
Excellent! Thank you! I'm happy to merge this.

tejas-ae:
Thanks!

PR #1377, opened 2026-07-31
Escape control characters in JSON string serialization

arisoik:
JSONParser.escapeString() didn't escape control characters. A mount error
message containing a newline produced invalid JSON in the
`/mount/get-config` response. The browser's `responseType='json'` XHR
silently turns that into `null`, so Mount.vue's poll loop threw trying to
read `.error` off `null` and retried forever instead of surfacing the actual
error — looked like an infinite "Mounting drive..." spinner with no
feedback.

Found this while debugging a mount failure on Linux that turned out to be
unrelated: Arch/CachyOS doesn't ship `gvfsd-dav` in the base `gvfs` package
(it's in `gvfs-dnssd` there; `gvfs-backends` on Debian/Ubuntu; already
included on Fedora) — a distro packaging issue, not something fixable here,
but worth knowing if "Location is not mountable" shows up again.

### DIFF
--- a/src/peergos/server/net/MountConfigHandler.java
+++ b/src/peergos/server/net/MountConfigHandler.java
@@ -318,10 +318,6 @@ public class MountConfigHandler implements HttpHandler {
         }
     }
 
-    private static boolean isLinux() {
-        return System.getProperty("os.name", "").toLowerCase().contains("linux");
-    }
-
     private static void openInFileExplorer(String mountPoint) throws IOException {
         // Android isn't handled here — the host app exposes a JS bridge
         // (MainActivity.openMountInFiles) that fires a SAF browse intent with the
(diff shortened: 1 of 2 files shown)


ianopolous:
linux mount works fine on ubuntu derivatives. Can we restrict this to Arch?

arisoik:
@ianopolous You're right! I re-tested with unmodified Digest auth once `gvfsd-dav` was actually present, and it works fine. The failures were entirely caused by Arch/CachyOS not shipping `gvfsd-dav` in the base `gvfs` package (it's in `gvfs-dnssd` there), not auth type. Reverted the Basic auth change. This PR is now just the JSON-escaping fix.

ianopolous (approved):

PR #1385, opened 2026-08-03
Fix AsyncReader dropping CBOR records that straddle a chunk boundary

arisoik:
## Summary
- `parseStreamRecurse` sized/bounded its read buffer using only the freshly-read byte count (`bytesRead`), ignoring a carried-over `prefix` from a prior partial read
- A CBOR record straddling a 5MiB chunk boundary could cause subsequent records to be silently dropped - affects friend-capability sharing records and social feed parsing once accumulated data exceeds ~5MiB
- The sibling method `parseLimitedStreamRecurse` already accounts for this correctly; this fix mirrors its buffer/offset handling

## Testing
- Added `CborObjects.parseStreamAcrossChunkBoundary` with object sizes chosen to deterministically straddle the boundary
- Confirmed it loses 2 of 3 objects on the pre-fix code and recovers all 3 with the fix

### DIFF
--- a/src/peergos/server/tests/CborObjects.java
+++ b/src/peergos/server/tests/CborObjects.java
@@ -3,8 +3,10 @@ package peergos.server.tests;
 import org.junit.*;
 import peergos.shared.cbor.*;
 import peergos.shared.io.ipfs.Multihash;
+import peergos.shared.user.fs.AsyncReader;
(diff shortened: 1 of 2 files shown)


ianopolous (approved):
Thank you!

PR #1386, opened 2026-08-03
Fix orphaned chunk leak when truncating a file into its first chunk

arisoik:
## Summary
- `FileWrapper.truncate()` skipped chunk deletion entirely whenever the new size fit within the first chunk, instead of just skipping deletion of chunk 0 itself
- The trailing chunks stayed reachable in the writer's CHAMP forever, and were never enumerated by the eventual file-delete path either, since chunk count is derived from the already-truncated size
- Fix: only skip deletion for genuinely single-chunk files; otherwise start deletion at chunk 1 (chunk 0 is the file's own entry point and must never be deleted here)

## Testing
- Extended `UserTests.truncate()` to assert the orphaned chunk is actually deleted after truncating a multi-chunk file down into its first chunk
- Confirmed the new assertion fails on the pre-fix code and passes with the fix

### DIFF
--- a/src/peergos/server/tests/UserTests.java
+++ b/src/peergos/server/tests/UserTests.java
@@ -1480,9 +1480,15 @@ public abstract class UserTests {
 
         // truncate to first chunk
         int truncateLength2 = 1 * 1024 * 1024;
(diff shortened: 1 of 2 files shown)


ianopolous (approved):
Good catch!

PR #1387, opened 2026-08-08
Fix sync dir delete failing when only ignored files remain

arisoik:
When a synced directory's real contents are deleted, the sync engine
tries to remove the now-supposedly-empty directory too. If it still
contains a file we deliberately never sync (e.g. .DS_Store, which
macOS Finder recreates automatically), the delete throws
`IllegalStateException: Trying to delete non empty directory`. That
exception wasn't caught, so it escaped `syncDir` and permanently
marked the sync pair as ERROR — every subsequent sync attempt threw
the same way, since the ignored file is never touched.

Fix: `LocalFileSystem.delete` now checks whether everything left in
a non-empty directory is in `IGNORED_FILENAMES` before giving up. If
so, it deletes those files too and proceeds with the directory
delete instead of throwing. `IGNORED_FILENAMES` is now
package-visible so `LocalFileSystem` can reference it directly.

Left untouched: `PeergosSyncFS`/remote deletes. `.DS_Store` is
filtered out of every upload path before it ever reaches the
remote, so this specific failure can't occur there under normal
use.

Test: `ignoredFileInDirIsCleanedUpOnDelete` reproduces the crash
(syncs a file, adds a `.DS_Store` locally, deletes the folder
remotely) and confirms the sync now completes cleanly, the
directory including the ignored file is gone from both sides, and a
second sync pass is stable.

### DIFF
--- a/src/peergos/server/sync/SyncFilesystem.java
+++ b/src/peergos/server/sync/SyncFilesystem.java
@@ -32,6 +32,11 @@ public interface SyncFilesystem {
 
     void bulkDelete(Path dir, Set<String> children);
 
+    /**
+     * @return the names of the children of dir, or empty if dir is absent or not a directory
+     */
+    Set<String> getChildNames(Path dir);
+
     void moveTo(Path src, Path target);
 
     long getLastModified(Path p);
--- a/src/peergos/server/sync/SyncFilesystem.java
+++ b/src/peergos/server/sync/SyncFilesystem.java
@@ -32,11 +32,6 @@ public interface SyncFilesystem {
 
     void bulkDelete(Path dir, Set<String> children);
 
-    /**
-     * @return the names of the children of dir, or empty if dir is absent or not a directory
-     */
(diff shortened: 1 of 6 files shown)


src/peergos/server/sync/SyncFilesystem.java
```diff
@@ -32,6 +32,11 @@ public interface SyncFilesystem {
 
     void bulkDelete(Path dir, Set<String> children);
 
+    /**
+     * @return the names of the children of dir, or empty if dir is absent or not a directory
+     */
+    Set<String> getChildNames(Path dir);
```
ianopolous: I don't think we need this. It would be cleaner to just check for the ignored files in LocalFileSystem.delete

arisoik: Simplified in https://github.com/Peergos/Peergos/pull/1387/commits/4478473e90791e19ec7914c95adaa8ce357b5343

ianopolous (approved):

PR #1388, opened 2026-08-08
Fix off-by-one padding the last chunk during prefetch

arisoik:
## Summary
Off-by-one in `LazyInputStreamCombiner`'s parallel prefetch: `totalChunks` is a chunk count, but the bound treated the last valid index as still "not the last chunk," so the final chunk was always sized to a full `Chunk.MAX_SIZE` and zero-padded past the real file length instead of using its true (shorter) length.

## Testing
New test uploads a file whose size isn't an exact multiple of `Chunk.MAX_SIZE`, reads through the buffered/prefetch path, and asserts no read past the true end of the file succeeds. Confirmed failing before the fix, passing after.


ianopolous:
The file block buffer is not valid because it needs to serve requests for the cid. Nonqwritethroughstorage is not used and so doesn't need tests. The 3rd fix on its own could be merged.

arisoik:
> The file block buffer is not valid because it needs to serve requests for the cid. Nonqwritethroughstorage is not used and so doesn't need tests. The 3rd fix on its own could be merged.

First two parts reverted in https://github.com/Peergos/Peergos/pull/1388/changes/08629092860fa1daa55091140a734614c5b459a8 and https://github.com/Peergos/Peergos/pull/1388/changes/94a5ed8f0323269ed26691d4056b964d104054b7.

ianopolous (approved):

PR #1399, opened 2026-08-16
Add pause to sync, and fix pair add/remove handling

arisoik:
Four independent bug fixes, then pause on top. Each commit stands alone.

## Fixes

**`get-pairs` could resurrect a removed folder.** It snapshots the config, then
asynchronously resolves every link over the network and writes the whole snapshot
back. Remove a folder while that is in flight and the pre-removal snapshot lands on
top of it. `updateRemotePaths` now resolves outside the lock and merges into the
config as it is at write time, and only writes when a path actually changed.
Reproducible by removing a pair while polling `get-pairs`: before, it reappeared
within ~2s every time.

**Overlapping pairs were accepted.** Syncing a folder that contains, or sits inside,
another synced folder makes each pair mirror the other's files and propagate its
deletions. `add-pair` now rejects either direction. Paths are compared with
separators normalised, and case-insensitively on Windows and macOS where the
filesystem treats them as the same folder.

**The sync endpoints accepted cross-site POSTs.** Any page the user visits could
pause syncing or remove a folder with a simple POST to localhost — no preflight, and
the caller cannot read the reply, but the side effect lands. Requests declaring a
different origin are now refused, mirroring the check `WebAuthnHandler` already
applies. Requests with no `Origin` header (native and app clients) are unaffected.

**Removing a folder stranded the others.** `remove-pair` cancels the running pass,
which abandons every folder it had not reached, and the runner then slept its full
interval. It now asks for another pass immediately. Two related fixes: a cancel no
longer leaks into the next pass (which used to consume it and abort before syncing
anything), and `runNow()` no longer interrupts mid-transfer — it records the request
so a changed config is picked up as soon as the pass unwinds.

Smaller ones: a re-added folder no longer reports the old pair's status; a pair's
stale error is cleared when its next attempt starts; progress names the path within
the synced folder instead of the leaf, since several subfolders can hold the same
filename; a status record with no message is no longer reported as one; and the
remote scan says what it is doing, as it rehashes anything it cannot match from
cache — minutes of silence on large files.

## Pause

New `pause` action, `sync-now` resumes, and `status` gains `paused`.

The flag is stored in `sync-config.json`, so a pause survives a restart. That matters
most on Android, where the process is killed routinely and an in-memory flag would
last minutes. Configs written by earlier versions have no key and are treated as
running.

Pausing stops the pass in flight, not just the schedule: fresh uploads, resumed
uploads and downloads all check for cancellation, so a transfer stops in about a
second instead of running to completion. A stop the user asked for is recorded as
stopped rather than as a failure, so a paused folder is not left looking broken.

## Testing

Exercised against a live account with two sync pairs: pause mid-transfer and resume;
remove a folder mid-upload and re-add it; break a folder and repair it; add
overlapping pairs in both directions; cross-origin requests refused while same-origin
and header-less ones pass; pause across a daemon restart; and a config written
without the new key still loads and syncs.

### DIFF
--- a/src/peergos/server/net/SyncConfigHandler.java
+++ b/src/peergos/server/net/SyncConfigHandler.java
@@ -39,7 +39,6 @@ import java.util.concurrent.ForkJoinPool;
 import java.util.function.Function;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 public class SyncConfigHandler implements HttpHandler {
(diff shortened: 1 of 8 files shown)


ianopolous (approved):
Nice!

Math1987:
I think there may be a small pause race at current master:

1. `ThreadBased` passes its top-of-loop `isPaused()` check.
2. `pause()` then sets both `paused` and `cancelled`.
3. `DirectorySync.syncDirs()` immediately calls `status.resume()`, clearing `cancelled`.
4. Since `isCancelled()` does not include `paused`, that pass can continue despite the pause.

Would you accept a small follow-up PR with a deterministic regression test and a change making `resume()` preserve cancellation while paused? `unpause()` already clears `paused` before calling `resume()`, so one-shot cancellation would keep its current behaviour.

arisoik:
@Math1987 indeed, this is a real race and your steps are right. Of course, you can create a PR otherwise I can handle it for you if you want, thank you for reporting it!

PR #1401, opened 2026-08-18
Keep sync pause active across pass reset

Math1987:
## Summary
- keep cancellation set while sync is paused
- add a deterministic regression test for the race between the runner pause check and pass startup

## Why
`ThreadBased` can pass its `isPaused()` check just before `pause()` sets both flags. `DirectorySync.syncDirs()` then calls `resume()`, which previously cleared `cancelled` and allowed that pass to continue.

`unpause()` still clears `paused` before calling `resume()`, so one-shot cancellation behavior is unchanged.

## Testing
- `ant dist` with Homebrew OpenJDK 26
- targeted pause-race regression: PASS

Follow-up to #1399.

### DIFF
--- a/src/peergos/server/sync/SyncRunner.java
+++ b/src/peergos/server/sync/SyncRunner.java
@@ -47,7 +47,8 @@ public interface SyncRunner {
         }
 
         public synchronized void resume() {
-            cancelled.set(false);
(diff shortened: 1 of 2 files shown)


ianopolous (approved):
Thanks!

PR #1403, opened 2026-08-21
Sync: report stops accurately and stop losing folder state

arisoik:
Fixes found while running the sync on Android and desktop.

- A pass that syncs a subset of pairs (only the ones allowed on mobile data)
  deleted the logs, status files and sync state databases of every pair it did
  not cover, forcing a full rescan. syncDirs now takes an everyPair flag and
  only tidies up when it really has them all.
- One link the server will not open aborted the whole pass; the other pairs
  now keep syncing and the failure names the folder it belongs to.
- A stopped pass reported its folders as SYNCED. Only a pause leaves progress
  on show now; any other stop says why it stopped.
- Bulk uploads and deletes ignored cancellation, so a stopped sync kept
  transferring until the batch finished.
- syncDirs returns whether it left a folder behind, so callers can retry.
- Failed requests are reported in words instead of the raw url, which told the
  user nothing and echoed their owner key back at them.
- The sync config is written to a temp file and moved into place, so a crash
  mid write cannot truncate it.
- The local scan logs its start like the remote one, so a long hash is not
  silence.

### DIFF
--- a/src/peergos/server/net/SyncConfigHandler.java
+++ b/src/peergos/server/net/SyncConfigHandler.java
@@ -34,6 +34,7 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystemException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
 import java.util.*;
 import java.util.concurrent.ForkJoinPool;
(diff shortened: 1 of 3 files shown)


ianopolous (approved):
All good fixes! Thank you.

PR #1414, opened 2026-08-28
Report transfer progress once per MiB

arisoik:
`LocalFileSystem.setBytes` reads in 4 KiB and reported progress on every read past the
first MiB, so a 5 MiB download emitted 1025 messages where 5 carry information. The upload
paths in `PeergosSyncFS` do the same per chunk, at about two messages per MiB.

Every one of those messages is appended to the pair log and written to the pair's status
file, which is a write plus an atomic rename. Measured on an Android device, one status
write costs ~3.7 ms, so a 100 MiB download spends ~95 s writing status alone, on the
download loop's critical path, and churns the 1 MiB pair log hard enough to lose its
history.

- one `SyncFilesystem.perMiB(verb, size, path, progress)` now decides when a figure is
  worth reporting, replacing three copies of the rule and of the message string
- `PairStatus.setError` and `setStatus(SyncStatus)` skip the write when the value is
  unchanged; a pass clears an already-clear error and re-sets the same state each time
  round. The message setter still writes every call, because the timestamp it stamps is
  what the view shows as "at <time>"

Tests: `progressIsReportedOncePerMiB` fails on master with 1025 messages for a 5 MiB file;
`redundantStatusWritesAreSkipped` compares the status file's inode, since a redundant write
writes back identical bytes. SyncTests, SyncRunnerTests and SyncWatcherTests pass.

Verified end to end on desktop: a 6 MiB file through a real sync pair logs exactly six
progress lines, and a 12 MiB upload on Android logs twelve, against 2.05 per MiB before.

### DIFF
--- a/src/peergos/server/sync/LocalFileSystem.java
+++ b/src/peergos/server/sync/LocalFileSystem.java
@@ -7,6 +7,7 @@ import peergos.server.simulation.FileAsyncReader;
 import peergos.shared.crypto.hash.Hasher;
 import peergos.shared.crypto.hash.PublicKeyHash;
 import peergos.shared.user.fs.*;
+import peergos.shared.util.ProgressConsumer;
 import peergos.shared.util.PathUtil;
 import peergos.shared.util.Triple;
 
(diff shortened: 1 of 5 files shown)


ianopolous (approved):
Nice

PR #1435, opened 2026-09-15
Stop the gc sweeping when a block could not be read

arisoik:
Supersedes the earlier version of this branch, which you correctly rejected: it let a user
stop their own collection by publishing a pointer to a block they never uploaded. That is
fixed here, and the s3 tests you asked for are what proved the first attempt was still
broken.

The bug

markReachable catches any error while walking, logs it, and returns true. Everything below
the failure stays unvisited, which at the sweep is indistinguishable from unreachable, so
deleteUnreachableBlocks deletes it. Production passes (cd, rd, c) -> Futures.of(true) as
deleteConfirm, so nothing downstream catches it, and gc runs every 12h.

getWithBackoff retries only RateLimitException, so an AccessDenied (rotated or expired
credentials), a corrupt cbor block, or a local sqlite error propagates on the first attempt.

Why 404s have to be separate

Telling the gc to stop on any error is exploitable: publishing a pointer to a block that was
never uploaded costs nothing and would stop that user's collection for good, while their
usage still falls. So absence and failure are now distinguishable at the s3 read:

    if (! notFound) {
        LOG.log(Level.SEVERE, cause, cause::getMessage);
        throw new IllegalStateException("Could not read block " + hash, cause);
    }
    throw new BlockAbsentException(hash);

Previously notFound was computed but used only for logging - both cases threw the same
IllegalStateException("Missing block"). FileContentAddressedStorage and AuthedStorage
returned an empty link list for an absent block, which is the other half of the same
problem, and now signal absence the same way.

The gc treats absence as normal and carries on; only a block that is there but unreadable
stops the sweep for that owner. No snapshot is written on that path, so the next run marks
again from scratch.

Tests

S3BlockAbsenceTests runs a real S3BlockStorage against LocalS3Server:
- aBlockS3NeverHadReportsAbsence - a 404 reports absence
- aBlockS3FailsToServeIsNotAbsence - a refusal does not
- gcCollectsGarbageOnS3 - control, so the rest is not vacuous
- aDanglingPointerDoesNotStopCollection - the malicious case
- s3FailingToServeALiveBlockStopsCollection - an unreadable live block stops the sweep

Reverting the 404 split fails the first and fourth; removing the gate fails the fifth.
GCTests adds the same two cases over the in memory storage, and correctMarkPhase now also
asserts a clean walk reports no failures.

LocalS3Server gains refuseRequestsFor(key) to inject a refusal. AccessDenied is used
deliberately: the codes in RETRY_S3_CODES are retried with a long backoff rather than
surfacing.

One existing test changed: versionedGC registered a usage root for a cid it never stored,
so marking could not resolve it. fullGC does the same and then updates usage to the real
root; versionedGC never did. It now keeps usage in step with the pointer, and all its own
assertions hold unchanged.


ianopolous:
We rely on the "missing block" exception text elsewhere

ianopolous (approved):

